PR #44, opened 2017-03-02
Problem: we need REST API call for fty-info

bstepankova:


jimklimov:
.gitignore not updated (new ecpp ~ cpp)

src/web/src/sysinfo.ecpp
```diff
@@ -581,7 +582,7 @@ sysinfo_plain (
     "\t}\n"
     "}";
 }
-
+
```
jimklimov: Why the extra spaces? :)

src/web/src/sysinfo.ecpp
```diff
@@ -50,7 +50,8 @@
 #include "asset_types.h"
 #include "utils.h"
 #include "git_details_override.h"
-
+#include "tntmlm.h"
```
jimklimov: Is this really needed in this file? I don't think sysinfo has any MLM chats, it serves pretty static info...

Makefile.am
```diff
@@ -1036,8 +1036,9 @@ BUILT_SOURCES_bios_web_la = \
                       src/web/src/my_profile.cpp \
  		      src/web/src/email_test.cpp \
 		      src/web/src/input_power_chain.cpp	\
-			  src/web/src/email_feedback.cpp \
-			  src/web/src/email_vote.cpp
+		      src/web/src/email_feedback.cpp \
+	              src/web/src/email_vote.cpp \
```
jimklimov: inconsistent space vs tabs

jimklimov:
Generally looks ok, the nits above can be addressed separately. Note to myself: Gotta do something about passing Travis with this repo :\

Merged PR so we have binaries to test.

PR #56, opened 2017-03-08
Problem: debug info improvement and correction

bstepankova:
Solution: done

### DIFF
--- a/src/web/src/info.ecpp
+++ b/src/web/src/info.ecpp
@@ -43,29 +43,29 @@ int
 get_rc_info (std::string &output)
 {
     output.clear();
-    
+     
     // connect to mlm client
     MlmClientPool::Ptr client = mlm_pool.get ();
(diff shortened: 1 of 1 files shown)


src/web/src/info.ecpp
```diff
@@ -43,29 +43,29 @@ int
...
 {
     output.clear();
-    
+     
     // connect to mlm client
     MlmClientPool::Ptr client = mlm_pool.get ();
     if (!client.getPointer ())
     {
-        log_critical ("sysinfo: mlm_pool.get () failed.");
+        log_critical ("info: mlm_pool.get () failed.");
```
jimklimov: Why "info"? If it is critical (per log function name, and apparently the priority that it is logged) and causes the error-exit of the function, it is not mere "info". Also, doesn't the log routine prefix the relevant tag (e.g. "CRITICAL") before the message?

PR #73, opened 2017-03-13
Problem: we don't want to use database id in rest api call

bstepankova:
Solution: name used instead

### DIFF
--- a/src/web/src/alert_list.ecpp
+++ b/src/web/src/alert_list.ecpp
@@ -102,7 +102,7 @@ UserInfo user;
     std::string checked_state;
     std::string checked_asset;
     std::string checked_recursive;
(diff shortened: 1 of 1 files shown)


src/web/src/alert_list.ecpp
```diff
@@ -133,12 +133,17 @@ UserInfo user;
                         std::string ("one of the following values [ true | false ]").c_str ());
             }
 
-        if (!asset.empty ()) {
-            http_errors_t errors;
-            if (!check_element_identifier ("asset", asset, element_id, errors)) {
-                http_die_error (errors);
+        if (!asset.empty ())
+        {            
+            if (!is_ok_name (asset.c_str ()))
```
vyskocilm: Do we need to call is_ok_name? cc @thalman

PR #87, opened 2017-03-15
Problem: CI build fails

aquette:
Solution: Try with the upstream CZMQ

### DIFF
--- a/ci_build.sh
+++ b/ci_build.sh
@@ -255,8 +255,8 @@ if [ "$BUILD_TYPE" == "default" ] || [ "$BUILD_TYPE" == "default-Werror" ] || [ 
         echo ""
         BASE_PWD=${PWD}
         echo "`date`: INFO: Building prerequisite 'czmq' from Git repository..." >&2
-        # $CI_TIME git clone --quiet --depth 1 -b v3.0.2 https://github.com/zeromq/czmq.git czmq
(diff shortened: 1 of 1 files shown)


jimklimov:
Note/FYI: the 3.x CZMQ is effectively abandoned for a couple of years by upstream, and our fork just contains a few commits on top of original `czmq-v3.0.2` to address some nits.

The real way forward is to build our whole project with current CZMQ4 (and also then using pre-compiled packages to speed up the builds), for which there are established `czmq4` branches in several of the 42ity repos where it mattered but otherwise almost no development traffic. Analysis during FOSDEM hackathon showed that there are about 3 routines whose usage patterns have changed, and a handful of locations to fix. Work done by @vyskocilm in https://github.com/42ity/fty-rest/commit/e28f664b8daba401ca042196fdcc57f497084d1c can be a pointer to the needed changes, or not... someone has to try and see :)

aquette:
Thx for the clarification @jimklimov.
But still, fty-rest builds are still failing with undef on zmq and I don't get the why. Any idea? I tried locally with ci_build, no deposit installed and gcc 4.8... can't reproduce

aquette:
@vyskocilm and @jimklimov : understood, but I still don't get the why builds are failing on *zmq undefs!
any clarification appreciated, thx.

PR #95, opened 2017-03-16
Problem: ID checking expects number but it is string

thalman:
Solution: Rewrite check, remove obsolete tests

Signed-off-by: Tomas Halman <TomasHalman@eaton.com>


jimklimov:
Comment for posterity: while it seems like some of the tests should have stayed (e.g. some patterns are now valid names, others might still be invalid), the main issue with the dropped tests is that they require a running database, so without an additional mockdb setup they are no longer "unit tests". So they went extinct.

PR #98, opened 2017-03-16
Problem: Created name is ugly and can have duplicities

thalman:
Solution: Create random unique name and update it later

Signed-off-by: Tomas Halman <TomasHalman@eaton.com>


aquette:
@thalman I may lack visibility but this seems not aligned with the story outcomes:
> Name for device : <element_sub-type>+MySQL seq number
> Example : ups-3, epdu-24
> The name is unique locally and never change.
or did I miss something? thx for any clarification on this

PR #124, opened 2017-03-18
Problem: fty-rest does not pass travis distcheck

jimklimov:
Solution: finally root out the bugs in source code and recipes, so travis tests for the repo pass green!

Consequence: from now on, red travis results for fty-rest is by 99% chance the fault of recent committers.


jimklimov:
Merged so I can reference this cleanly from 42ity/FTY

aquette:
Deserved kudos @jimklimov!

PR #125, opened 2017-03-18
[DO NOT MERGE YET] Use pkgconfig deps when building fty-rest

jimklimov:
While this PR ensures better (more correct) way of building, it will likely cause package builds to fail until our farm is set up to provide the expected packages from our github forks rather than (patched or vanilla) upstreams. Which should be done anyway, as part of guaranteeing FOSS compliance (that we do publish source to binaries we provide - and building from same sources is just simpler than other approaches, both for builds and the guarantees). This quest would likely be tackled in some next sprint.


jimklimov:
Part of this PR was split off into #136 that is "safe" (OS-image-wise) to merge already.

PR #131, opened 2017-03-20
Problem: input_power_chain call accepts  only numerical ids

bstepankova:
Solution: characters are allowed too

### DIFF
--- a/src/web/tntnet.xml
+++ b/src/web/tntnet.xml
@@ -356,7 +356,7 @@
     <mapping>
       <target>input_power_chain@bios_web</target>
       <method>GET</method>
-      <url>^/api/v1/topology/input_power_chain/([0-9]+)$</url>
+      <url>^/api/v1/topology/input_power_chain/.*$</url>
       <args>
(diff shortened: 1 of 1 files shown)


src/web/tntnet.xml
```diff
@@ -356,7 +356,7 @@
     <mapping>
       <target>input_power_chain@bios_web</target>
       <method>GET</method>
-      <url>^/api/v1/topology/input_power_chain/([0-9]+)$</url>
+      <url>^/api/v1/topology/input_power_chain/.*$</url>
```
jimklimov: WRONG FIX

1) Must be in parentheses (`$1` used below)
2) If we need this not-empty, `.+` at least
3) Do we really want ANY chars? Or same as in CI `[-.a-z0-9]`?

jimklimov: Otherwise, good catch ;)

jimklimov: Otherwise, good catch ;) :+1:

PR #137, opened 2017-03-21
 	Problem: incorrect error message

bstepankova:
Solution: error message corrected

### DIFF
--- a/src/web/src/average.ecpp
+++ b/src/web/src/average.ecpp
@@ -238,7 +238,7 @@ UserInfo user;
             if (streq (frame, "BAD_REQUEST")) {
                 zstr_free (&frame);
                 std::string die = "Data for type = '" + checked_type + "', step = '" + checked_step +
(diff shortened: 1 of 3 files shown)


src/web/src/average.ecpp
```diff
@@ -238,7 +238,7 @@ UserInfo user;
             if (streq (frame, "BAD_REQUEST")) {
                 zstr_free (&frame);
                 std::string die = "Data for type = '" + checked_type + "', step = '" + checked_step +
-                "', source = '" + checked_source + "', element_id = '" + std::to_string (checked_element_id) + "' ";
+                "', source = '" + checked_source + "', element_name = " + " '" + element_name + "'.";
```
jimklimov: `"', element_name = " + " '"` => `"', element_name = '"`

jimklimov: So now there's two spaces after `=`, and also a newly introduced period in the end of sentence.

PR #151, opened 2017-03-27
Problem: Cppcheck reports number of issues

vyskocilm:
Solution: this is longer term private cleanup, but easier now with a fewer libprivs. This eliminates some dead code and unused variables. The next step is to move all header files to one location and simplify the number of -I paths we have in the project. cppcheck reports so many false positives now.


jimklimov:
Looks good

PR #154, opened 2017-03-29
Problem: code structure of fty-rest is complicated

vyskocilm:
Solution: all c .cc files to src/shared, all .h files to src/include. src/db and src/persist was not touched yet.


jimklimov:
The idea looks great, hope the build ends up green :)

PR #161, opened 2017-04-04
Problem: asset parameter is optional

vyskocilm:
Solution: fail only if checked_asset is not empty

Signed-off-by: Michal Vyskocil <MichalVyskocil@eaton.com>

### DIFF
--- a/src/web/src/alert_list.ecpp
+++ b/src/web/src/alert_list.ecpp
@@ -153,10 +153,9 @@ UserInfo user;
 
     {
         int64_t dbid =  persist::name_to_asset_id (checked_asset);
-        if (dbid == -1) {
+        if (!checked_asset.empty () && dbid == -1) {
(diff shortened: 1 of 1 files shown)


src/web/src/alert_list.ecpp
```diff
@@ -153,10 +153,9 @@ UserInfo user;
 
     {
         int64_t dbid =  persist::name_to_asset_id (checked_asset);
-        if (dbid == -1) {
+        if (!checked_asset.empty () && dbid == -1) {
             http_die ("element-not-found", checked_asset.c_str ());
         }
-
         element_id = (uint32_t) dbid;
```
jimklimov: Just in case: is it acceptable for us to continue with `dbid == -1` (and so casted to `element_id = (uint32_t)(2^32-1)` when asset IS empty)?

PR #187, opened 2017-04-20
Problem: typo in log_debug

bstepankova:


src/db/topology2.cc
```diff
@@ -193,7 +193,7 @@ topology2_groups (
                 "group"
                 });
     }
-    log_debug ("group.size ()=%zu", );
```
jimklimov: Noted to @bstepankova : printing this processing of `%zu` would likely reach into the unknown - no vararg uint follows here, so this logging is likely to segfault.

PR #190, opened 2017-04-24
Problem: devices in group are not ordered

vyskocilm:
Solution: add order by there


vyskocilm:
syntax error in SQL, closing

PR #222, opened 2017-05-15
Problem: webserver and admin user allowed to call /bin/systemctl for anything

jimklimov:
Solution: secure the system back by using the wrapper script (via PATH envvar)

NOTE: Requires to be merged along with https://github.com/42ity/fty-core/pull/91 for REST API to still function

### DIFF
--- a/src/web/src/systemctl.ecpp
+++ b/src/web/src/systemctl.ecpp
@@ -20,6 +20,7 @@
 /*!
  * \file systemctl.ecpp
  * \author Karol Hrdina <KarolHrdina@Eaton.com>
+ * \author Jim Klimov <EvgenyKlimov@Eaton.com>
  * \brief  Provide Users ability to manipulate with some specified services
  */
  #><%pre>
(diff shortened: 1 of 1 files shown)


jimklimov:
Per discussion with @karolhrdina this security fix (and respective testing of performance/stability impact) belongs in a separate user story, not in the middle of his current work. SUDO rights needed for REST API were restored in https://github.com/42ity/fty-core/pull/92 and this PR is split into "passive" comment-commits https://github.com/42ity/fty-rest/pull/223 that can be merged quickly and an "active" change that is marked as DO NOT MERGE YET.

jimklimov:
Bumped the branch to resync with current master and solve merge conflicts.

jimklimov:
Bumped the branch to bring this PR in line with recently zproject'ized codebase of the component.

Also note that after this PR gets merged and tested, the PR https://github.com/42ity/fty-core/pull/92 (allowing webserver to call `/bin/systemctl`) can be reverted.

jimklimov:
@geraldguillaume @aquette : since our recent hot topic is improvement of security, I'd like to ask about this long-hanging PR. Its goal is to use our `systemctl` wrapper instead of the system binary, so limiting the impact of services that the webserver may manage.

While the codebase change is simple, its repercussions would want some testing so the system overall still works to the extent we need of it (e.g. make sure all modern services are listed in the wrapper, later to move into its configurability support, so the intended REST API operations do still happen). This QA was the main reason it was not merged a year ago; now with better test coverage we might just see quickly if it works.

jimklimov:
Rebased over current master branch state to avoid conflicts.

jimklimov:
Rebased over current master branch state to avoid conflicts, again.

jimklimov:
@geraldguillaume : as far as I am concerned, the consideration as a cautious OK remains for over a year now :) This change has a potential to break some tests, which should not be hard to fix, while adding security to elevated systems management calls by having them filtered same as for command line.

PR #240, opened 2017-06-14
Problem: We need REST API call to run scan.

jana-rapava:


src/web/src/scan_run.ecpp
```diff
@@ -0,0 +1,127 @@
...
+    else
+    if (rv == -3)
+        http_die ("internal-error", "client->recv () failed.");
+    else
+    if (rv == -4)
+        http_die ("request-param-required", "<range>");
+    else
+    if (rv == -5)
+        http_die ("internal-error", "fty-discovery returned malformed or unexpected message.");
+    else
```
jimklimov: ..else - what? Is there any return/report for success? :)

PR #252, opened 2017-06-29
Problem: GPIO REST API call is not working.

jana-rapava:


src/web/src/list_gpio.ecpp
```diff
@@ -138,9 +154,6 @@ get_list (std::string &output)
         zmsg_destroy(&resp);
         return -5;
     }
-
-    zmsg_destroy(&resp);
```
jimklimov: What happens `if (!result)` ? Shouldn't the removed destroy and return remain in place?

PR #254, opened 2017-07-03
Problem: Warning and exception should be addressed 

aquette:
Solution: Do it


jimklimov:
Just in case/sanity check: why is it not an even logged error, maybe higher in call stack, that `asset_id == 0` anymore? Alternately, perhaps test input at beginning of routine to bail out (or even assert) quickly, and go to SQL with presumed valid inputs?

PR #258, opened 2017-07-07
[DO NOT MERGE YET] Problem: Need to rework DNS handling for DHCP

aquette:
Solution: Also dump nameservers to /etc/udhcpc/dns
Note that there is a complementary PR in bios-boot -> src/etc/udhcpc.script
for this to take effect


aquette:
@geraldguillaume not there are remaining printf, to be at least removed prior to merging

jimklimov:
Error checking missing when you open(FILE). Indeed, printf should go or give way to log_debug.

PR #259, opened 2017-07-12
Problem: We need to be able to pass a multiscan on JSON format

barraudl:
Problem: We need to be able to pass a multiscan on JSON format

### DIFF
--- a/src/web/src/scan_run.ecpp
+++ b/src/web/src/scan_run.ecpp
@@ -27,6 +27,7 @@
 #include "tntmlm.h"
 #include "helpers.h"
 #include "utils_web.h"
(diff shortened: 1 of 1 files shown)


src/web/src/scan_run.ecpp
```diff
@@ -169,6 +213,67 @@ run_localscan ()
 }
 
 int
+run_multiscan(std::string json)
+{
+     // connect to mlm client
+    MlmClientPool::Ptr client = mlm_pool.get ();
+    if (!client.getPointer ())
+    {
```
MimiXav: be coherent with bracket with the remaining code

PR #265, opened 2017-07-20
Problem : I want more log when new db credentials are set

geraldguillaume:
Sometime, when accepting the licence, tntnet fail to connect to DB because of bad credentials.
To investigate this issue
- add [pid.thead_id] in our log
- add info log "Updating db url with DB_USER=" 
- force read bios-db-rw file even if it already exist

### DIFF
--- a/src/shared/dbpath.cc
+++ b/src/shared/dbpath.cc
@@ -61,7 +61,7 @@ s_dropdq (char* buffer) {
 
 // read /etc/default/bios-db-rw and update url global variable
 // @return 0 if OK, -1 if KO
-bool dbreadcredentails(){
+bool dbreadcredentials(){
     if(!shared::is_file (PASSWD_FILE))return false;
     // and setup db username/password
     log_debug("Reading %s ..",PASSWD_FILE);
--- a/src/shared/dbpath.cc
+++ b/src/shared/dbpath.cc
@@ -19,8 +19,14 @@
  */
 
 #include "dbpath.h"
+#include "log.h"
 
 #include <stdlib.h>
+#include <stdio.h>
+#include <fstream>
+#include <string.h>
+#include <stdlib.h>
+#include "filesystem.h"
 
 std::string url = std::string("mysql:db=box_utf8;user=") +
               ((getenv("DB_USER")   == NULL) ? "root" : getenv("DB_USER")) +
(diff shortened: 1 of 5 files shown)


src/shared/dbpath.cc
```diff
@@ -19,18 +19,65 @@
...
+    memset (db_passwd, '\0', 256);
+    dbpasswd.getline (db_passwd, 256);
+    s_dropdq (db_passwd);
+    dbpasswd.close ();
+    if(db_user==NULL || db_passwd==NULL)return false;
+    putenv (db_user);
+    putenv (db_passwd);
+    dbpath();
+    return true;
+}
```
jimklimov: No newline at EOF :)

src/shared/dbpath.cc
```diff
@@ -19,18 +19,65 @@
...
+#include "log.h"
 
 #include <stdlib.h>
+#include <stdio.h>
+#include <fstream>
+#include <string.h>
+#include <stdlib.h>
+#include "filesystem.h"
 
 std::string url = std::string("mysql:db=box_utf8;user=") +
```
jimklimov: While we are here, perhaps refactor to use a single string-generation routine (or macro) to initialize and re-set the `url` with same code? Copies like these need extra revision to make sure they do the same thing...

src/shared/dbpath.cc
```diff
@@ -19,18 +19,65 @@
...
+        while (*dq_ptr) {
+            *dq_ptr = *(dq_ptr+1);
+            dq_ptr++;
+        }
+    }
+}
+
+// read /etc/default/bios-db-rw and update url global variable
+// @return 0 if OK, -1 if KO
+bool dbreadcredentails(){
```
jimklimov: credentIALs

src/web/src/license_POST.ecpp
```diff
@@ -125,24 +108,12 @@ UserInfo user;
...
-        dbpasswd.getline (db_passwd, 256);
-        s_dropdq (db_passwd);
-        dbpasswd.close ();
-
-        putenv (db_user);
-        putenv (db_passwd);
-        dbpath ();
+    }else{
+        //enforce reload credential
+        if (!dbreadcredentails())
```
jimklimov: If I get this changeset right, then beside relocating some code, this is the one effective logical change so far?

So now we create the license file during license_POST, and if a DB credential file already exists - re-read it (just in case it was not there or up-to-date when tntnet initially started), right?

jimklimov:
Not sure why, but this build upsets `test-log` run on Travis CI. Previous builds in history were green...

PR #275, opened 2017-08-10
[WIP - DO NOT MERGE YET] Problem: REST API backend can call database while it is not ready yet

EldoreiJK:
Solution: Added check for database being ready to rest calls.


src/web/src/helpers.cc
```diff
@@ -419,6 +419,19 @@ check_user_permissions (
...
 
 char*
+get_current_db_initialized_file (void)
+{
+    char *current_db_initalized_file = NULL;
+    char *env = getenv (EV_DB_INITIALIZED_DIR);
+
+    int rv = asprintf (&current_db_initalized_file, "%s/fty-db-ready", env ? env : "/var/run/");
+    if ( rv == -1 ) {
+        return NULL;
```
jimklimov: I wonder if we should also check `if (current_db_initalized_file) free(current_db_initalized_file )` here just in case?

src/web/src/helpers.cc
```diff
@@ -419,6 +419,19 @@ check_user_permissions (
 }
 
 char*
+get_current_db_initialized_file (void)
+{
+    char *current_db_initalized_file = NULL;
+    char *env = getenv (EV_DB_INITIALIZED_DIR);
+
+    int rv = asprintf (&current_db_initalized_file, "%s/fty-db-ready", env ? env : "/var/run/");
```
jimklimov: Trailing slash is superfluous here.... but should be harmless.

src/web/src/server_status.ecpp
```diff
@@ -0,0 +1,47 @@
...
+</%pre>
+<%request scope="global">
+bool database_ready;
+</%request>
+<%cpp>
+{
+    /* Check whether database has started and initialized yet */
+    struct stat db_stat;
+    char *database_ready_file = get_current_db_initialized_file();
+    log_debug("database_ready_file='%s'", database_ready_file);
```
jimklimov: No check here and below for `NULL` values of `database_ready_file`... did you check how well `stat()` handles that?

src/web/src/server_status.ecpp
```diff
@@ -0,0 +1,47 @@
...
+    database_ready = stat(database_ready_file, &db_stat);
+    free (database_ready_file); database_ready_file = NULL;
```
jimklimov: From `man -s 2 stat` : 
````
RETURN VALUE
       On success, zero is returned.  On error, -1 is returned, and errno is set appropriately.
````

1) We might want to print/log the `perror()` or at least the `errno` if we got non-zero
2) The Boolean "true" would be having this non-zero output which is error and inverse of what we want by this variable's name and usage as
````
   if (!database_ready) { log_debug ("Database is not ready yet.");  ...
````

EldoreiJK:
OK, comments are addressed now.

jimklimov:
Thanks, looks good now :) :+1:

PR #276, opened 2017-08-11
Fixing REST API tests.

jana-rapava:


jimklimov:
It seems that some places check for `rv < 0` meaning errors (with `>=0` being good outputs like a size), and others mean their `rv==0` as good and `rv!=0` as errors. This at least reads inconsistently,  maybe better use `rv<0` everywhere?

Also, a few places get the `rv` and do not check or return it further - which is both a compiler warning and a way to lose the known failures.

And at least once a check that culminated in `bios_throw` on empty/bad input became a `Database error` which may not be what we want (alone) there - but I did not research the wider context.

Hopefully the sane and insane inputs from CI will show if any of these suspicions are actual problems or none are, but devs might want to take a second look in advance ;)

PR #279, opened 2017-08-14
Problem: Database failure is not detected in asset import/export.

jana-rapava:
Solution: Add error-checking code.


src/db/inout/exportcsv.cc
```diff
@@ -314,6 +330,8 @@ void
             else {
                 std::string extname;
                 int rv = persist::name_to_extname (groups[i], extname);
+        if (rv != 0)
```
jimklimov: Indentation :)

PR #281, opened 2017-08-14
Problem: We need to check for DB failure in asset export/import.

jana-rapava:


src/db/inout/importcsv.cc
```diff
@@ -513,6 +530,15 @@ static std::pair<db_a_elmnt_t, persist::asset_operation>
             auto ret = select_asset_element_by_name
                 (conn, value.c_str());
             if ( ret.status == 0 ) {
+                if (ret.errsubtype == DB_ERROR_NOTFOUND) {
+                    log_info ("logical_asset '%s' does not present in DB, rejected",
```
jimklimov: is not ;)
and more tabs before next line (value.cstr)

jimklimov:
Looks great, waiting for Travis

PR #289, opened 2017-08-17
Problem: GPIO port is hard coded

bstepankova:
Solution: variable created


EldoreiJK:
I don't really like adding "rackcontroller" to "types", please investigate whether it is really needed or not.

PR #301, opened 2017-09-06
Problem: unsigned long is not always uint64_t

jimklimov:
Solution: change function signature


jimklimov:
This is a probable cause of recent failures on ARM. Please wait for Travis to pass on X86 at least before merging ;)

PR #308, opened 2017-09-27
Problem : Fix uuid memory leak 

geraldguillaume:
-> uuid should be destroyed after usage
-> reduce verbosity of security headers


MimiXav:
Ok for me

PR #312, opened 2017-10-10
[DONT MERGE YET] Problem: csv import gets broken by rackcontroller-0

EldoreiJK:
Solution: Added support for rackcontroller-0 to CSV import.

### DIFF
--- a/src/db/dbhelpers.h
+++ b/src/db/dbhelpers.h
@@ -53,59 +53,59 @@ typedef std::function<void(const tntdb::Row&)> row_cb_f ;
 
 template <typename T>
 struct db_reply{
-    int status; // ok/fail
-    int errtype;
-    int errsubtype;
-    uint64_t rowid;  // insert/update or http error code if status == 0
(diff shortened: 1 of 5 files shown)


EldoreiJK:
Still need some tests, and I'm requesting your comments.

src/db/dbhelpers.h
```diff
@@ -53,59 +53,59 @@ typedef std::function<void(const tntdb::Row&)> row_cb_f ;
 
 template <typename T>
 struct db_reply{
-    int status; // ok/fail
```
jimklimov: It seems that the changes in `dbhelpers.h` are all about breaking indentation, and nothing functional. Can you pluck this one out of the PR please?

jimklimov:
Probably `libmlm-dev` needs to be uncommented in `.travis.yml`.

EldoreiJK:
Good point about useless changes, reverted both 2 files with no more changes than whitespaces.

PR #317, opened 2017-10-30
 Problem: GET /admin/licensing returns wrong Content-Type.

jana-rapava:


src/web/src/json.ecpp
```diff
@@ -57,20 +57,16 @@
...
-	static cxxtools::Regex rex_GET_text_csv ("^/api/v1/asset/export$");
-	static cxxtools::Regex rex_GET_undecided("^/api/v1/admin/getlog/.*$");
+	static cxxtools::Regex rex_license("^/api/v1/admin/license$");
+	static cxxtools::Regex rex_export ("^/api/v1/asset/export.*$");
+	static cxxtools::Regex rex_licensing ("^/api/v1/admin/licensing$");
+	static cxxtools::Regex rex_getlog ("^/api/v1/admin/getlog/.*$");
 
 	if(request.getMethod() == "GET") {
 		log_debug("%s", ("GetURL==" + request.getUrl()).c_str() );
+        std::string url = request.getUrl ();
```
jimklimov: indent

jimklimov:
Per https://github.com/42ity/fty-rest/blob/master/src/include/utils_web.h#L147 `http_die()` would set the content type if not yet set, so LGTM.

PR #323, opened 2017-11-13
Problem: read-only HW based values may be overwritten by CSV import for RC-0, CSV import fails on invalid date format with error in UI

EldoreiJK:
Solution: removed HW based read-only attributes from CSV import for RC-0


EldoreiJK:
Solution: CSV rows that can't be imported are not considered for ids check, thus don't generate "duplicate" error

PR #324, opened 2017-11-14
Problem: we need new subtype 'gpo'

bstepankova:
Solution: added


aquette:
side note on this commit: we should definitely get this enum and strings from fty-asset, instead of duplicating things. And preferably move to some common part

PR #334, opened 2017-12-07
topology2: Initialize the asset_order for the root asset

michal42:
Signed-off-by: Michal Marek <MichalMarek1@eaton.com>


michal42:
wrong target

PR #336, opened 2017-12-08
topology2: Initialize the asset_order for the root asset

michal42:
The 1.3 build setup is discarded in this merge

### DIFF
--- a/src/db/topology2.cc
+++ b/src/db/topology2.cc
@@ -660,6 +660,7 @@ topology2_from_json_recursive (
                 it2.name = s_get (row, NAME),
                 it2.subtype =  from_subtype;
                 it2.type =  from_type;
+                it2.asset_order = 0;
             }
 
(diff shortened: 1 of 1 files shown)


jimklimov:
I think there's too many commits speaking about "1.3" going into the master branch here. Development on master should not have any knowledge of 1.3 release and branding IMHO.

@aquette : I think this is one of the points we discussed about the problems we can get if doing development the other way around (developing in release branch, then merging it all back to the main development master branch), beside the master builds being backed by more resources and layers of testing. It "should" be fine to develop in one branch, then cherry-pick the set of commits to the other if applicable, e.g. with `git log --left-right --graph --cherry-pick --oneline master...release/IPM_Infra-1.3` or `git cherry release/IPM_Infra-1.3 master`...

michal42:
Replaced with https://github.com/42ity/fty-rest/pull/338

PR #342, opened 2017-12-14
Problem: periodic poller is based on mlm poller function that expires only when there are no messages

EldoreiJK:
Solution: separated poller from timer for periodic polling


EldoreiJK:
please squash/rebase and merge if good

PR #345, opened 2018-01-05
[DO NOT MERGE YET] Use pkgconfig deps when building fty-rest

jimklimov:
Remake of #125 ported to current state of codebase

Problem: our forks of some libraries now provide pkg-config metadata
    
Solution: use it (requires proper setup of packaging farm too)
* Makefile.am configure.ac : wrap the custom libraries we need with pkg-config detected data, not explicit `-lNAME` where possible
* Detect and configure support for our libmagic clone via pkgconfig as well - or fall back to no pkg-config with older GA packages
* Makefile.am : comment about libmagic without pkgconfig is no longer relevant
* ci_build.sh : improve debugging review a bit by showing the actual pkgconfig files for these packages


jimklimov:
This seems irrelevant now after the switch to zproject (and earlier updating our 3rd-party forks to have the pkg-config files packaged).

PR #346, opened 2018-01-17
Ipmcvg 1412 improve alert system

nicodav:
Implement the sse protocol for send new Alert to the UI with a new eccp file : admin_sse.ecpp.
Base url : <host>/api/v1/sse?datacenter=xxx
Parameter : datacenter : the id of the datacenter to manage. 

Functionnality : 
-Must have a valid connection to get a sse connection.
-Subscribe to the ALERT stream and when receive an alert, send this alert via sse if the asset of the alert is a part of the datacenter.
-If no alert is send to UI after waiting 10 seconds, a heartbeat message is send to control the UI is still connected.
-Every minute, a message with the remaining time of the session is send.
-If the session is revoked or expired, the  sse connection is shutdown.
-If any notification (alert or heartbeat or session time) failed when send, the sse connection is shutdown
-The sse timout connection is the half of the tntnet request timeout

See https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events for sse specifications and recommandations

### DIFF
--- a/src/web/src/admin_sse.ecpp
+++ b/src/web/src/admin_sse.ecpp
@@ -19,8 +19,8 @@
  #><#
 /*!
  \file alert_list.ecpp
- \author Karol Hrdina <KarolHrdina@Eaton.com>
- \brief Retrieves a list of alerts known to the system according specified parameters
+ \author nicolas DAVIET <nicolasdaviet@Eaton.com>
+ \brief Open an sse connection to send any notification to UI
 */
 #><%pre>
 #include <exception>
--- a/src/web/src/admin_sse.ecpp
+++ b/src/web/src/admin_sse.ecpp
@@ -18,7 +18,7 @@
  #
  #><#
 /*!
- \file alert_list.ecpp
+ \file admin_sse.ecpp
  \author nicolas DAVIET <nicolasdaviet@Eaton.com>
  \brief Open an sse connection to send any notification to UI
 */
@@ -44,10 +44,6 @@
 #include "db/assets.h"
 #include "cleanup.h"
 #include "helpers.h"
-
-#define RFC_ALERTS_LIST "rfc-alerts-list"
-
-
 
 </%pre>
 <%request scope="global">
@@ -75,8 +71,6 @@ bool database_ready;
     std::string dc = qparam.param("datacenter");
     log_debug ("Request parameters - Initial tainted values received: datacenter = '%s'\n",
             dc.c_str ());
-
-
 
     if (!dc.empty ())
     {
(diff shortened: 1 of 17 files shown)


src/web/src/admin_sse.ecpp
```diff
@@ -0,0 +1,334 @@
...
+    if (!dc.empty ())
+    {
+        if (!is_ok_name (dc.c_str ()))
+        {
+            http_die ("request-param-bad", dc.c_str ());
+        }
+    } 
+
+    checked_dc = dc;
+    std::string checked_state = "ALL";
```
geraldguillaume: checked_state is never used

src/web/src/admin_sse.ecpp
```diff
@@ -0,0 +1,334 @@
...
+    log_debug ("Asset element id = '%" PRIu32 "'.", element_id);
+
+    if (!checked_dc.empty ()) {
+        auto asset_element = persist::select_asset_element_web_byId (connection, element_id);
+        if (asset_element.status != 1)
+            http_die ("internal-error", asset_element.msg.c_str ());
+        log_debug ("Asset element name = '%s'.", asset_element.item.name.c_str ());
+        desired_elements.emplace (std::make_pair (asset_element.item.name, 5));
+
+        if (checked_recursive.compare ("true") == 0) {
```
geraldguillaume: checked_recursive is initialized by "true" and never re-affected.
This variable looks useless

src/web/src/tokens.cc
```diff
@@ -253,11 +253,15 @@ BiosProfile tokens::verify_token(const std::string token, long int* uid, long in
     if (gid)
         *gid = l_gid;
 
+    time_t now = mono_time (NULL);
```
geraldguillaume: reuse now in the condition line bellow "if(mono_time(NULL).."

src/web/src/tokens.cc
```diff
@@ -253,11 +253,15 @@ BiosProfile tokens::verify_token(const std::string token, long int* uid, long in
...
     if (mono_time (NULL) > tme) {
         log_info ("verify_token: expired token for uid/gid %ld/%ld, authentication failed!", l_uid, l_gid);
         return BiosProfile::Anonymous;
     }
-
+    *expInSec = tme - now;
+    log_debug ("verify_token: expInSec: %ld",*expInSec);
```
geraldguillaume: This debug could really spam the system. I recommend to remove it

src/web/src/tokens.cc
```diff
@@ -253,11 +253,15 @@ BiosProfile tokens::verify_token(const std::string token, long int* uid, long in
...
+
```
geraldguillaume: two empty line is too much

src/web/src/admin_sse.ecpp
```diff
@@ -0,0 +1,334 @@
...
+
+    if (!dc.empty ())
+    {
+        if (!is_ok_name (dc.c_str ()))
+        {
+            http_die ("request-param-bad", dc.c_str ());
+        }
+    } 
+
+    checked_dc = dc;
```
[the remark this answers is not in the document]
nicodav: Remove checked_dc variable

PR #348, opened 2018-02-01
Problem: font files with Cache-Control HTTP header defined break IE

geraldguillaume:
Cache-Control and Pragma HTTP headers does not allow IE to download well font files resource on F5 or C^+F5  button refresh.
 
Solution : implement a new tntnet servlet "security_removeheaders" witch remove specific HTTP header for specific url mapping.
 

### DIFF
--- a/src/web/src/security_removeheaders.ecpp
+++ b/src/web/src/security_removeheaders.ecpp
@@ -1,6 +1,6 @@
 <#
  #
- # Copyright (C) 2017 Eaton
+ # Copyright (C) 2018 Eaton
  #
  # This program is free software; you can redistribute it and/or modify
  # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


src/web/src/security_removeheaders.ecpp
```diff
@@ -0,0 +1,48 @@
...
+ *     <target>security_removeheaders@bios</target>
+ *     <url>\.ttf$</url>
+ *     <args>
+ *         <remove>Cache-Control</remove>
+ *     </args>
+ *   </mapping>
+ * \author Gerald Guillaume <GeraldGuillaume@Eaton.com>
+ * \brief  remove HTTP headers defined as args in tntnet.xml
+ */
+#><%pre>
```
misterk72: Why do you need to comment this line? Isn't the <%pre> needed by the </%pre> 2 lines below?

geraldguillaume:
Genepi team validated it, so looks good of all

PR #349, opened 2018-02-07
 Problem: We need REST API call providing GPIO port counts to UI.

jana-rapava:


src/web/src/hw_capability.ecpp
```diff
@@ -0,0 +1,161 @@
...
+    MlmClientPool::Ptr client = mlm_pool.get ();
+    if (!client.getPointer ())
+    {
+        log_critical ("hw_capability: mlm_pool.get () failed.");
+        return -1;
+    }
+
+    zuuid_t *uuid = zuuid_new ();
+    zmsg_t *send = req_hw_cap (uuid, type);
+    int rv = client->sendto ("fty-info", "hwcapabaility", 1000, &send);
```
bstepankova: I think subject is supposed to be 'info' not 'hwcapabilities'

PR #350, opened 2018-02-08
Fix constexpr error in utils_web.h

michal42:
String literals are not guaranteed to be merged by the compiler, so
the result of "foo" == "foo" is undefined and thus not a constexpr.
Older gcc did not catch this, but gcc7 now rightfully errors out. As a
band-aid, change it to use strcmp(), which is not required by the
standard to be a constexpr for string constants, but it is a constexpr
with gcc. The proper fix would be to use an enum for error codes instead
of strings.

### DIFF
--- a/src/include/utils_web.h
+++ b/src/include/utils_web.h
@@ -94,7 +94,7 @@ constexpr ssize_t
 _die_idx(const char* key)
 {
     static_assert(std::tuple_size<_WSErrors>::value > N, "_die_idx asked for too big N");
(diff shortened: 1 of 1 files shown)


michal42:
Travis issue:

$ sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install coreutils git asciidoc xmlto libsasl2-2 libsasl2-dev sasl2-bin libmlm-dev
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package libmlm-dev
apt-get.diagnostics
apt-get install failed

PR #351, opened 2018-02-14
Switch fty-rest on zproject

barraudl:
Buildable, usable, but :

TO DO :
config.h and git_details_override.h/.cc no more generated (currently hard coded).
no more testing ("test" dirrectory was not take into consideration yet)
same for "tools" dirrectory

project need to be cleaned.

must have closer look on install process.
Build, usable, but :

TO DO :
config.h and git_details_override.h/.cc no more generated (currently hard coded).
no more testing ("test" dirrectory was not take into consideration yet)
same for "tools" dirrectory

project need to be cleaned.

must have closer look on install process.


jimklimov:
Generally a useful endeavor ;)

I did not get to scrap my old build recipes in favor of the new ones, so am curious how this pans out. A few notes:

* as seen in some other `*-rest` components, we have to jump through strange superficial hoops to support `*.ecpp` sources of "zproject classes" to be converted into `*.cc`/`*.cpp` and finally compiled. I thinks it would be quite beneficial to add direct zproject support for such use-case (e.g. another option for a class in `project.xml`), port our recipe code that was added manually in those components, and use this new zproject feature directly.
* From PR overview, I see that a lot of files (headers and sources of various helpers) were just removed, or so it seems from `git status`... is that expected? was the code ported into some other files?
* No really string opinion about the git_details files (especially since after the split of "core" into "fty-core" and "fty-rest" we have two conflicting copies). Historically one rationale in favor of them was that it lets us quickly evaluate (e.g. during tests) the source repo, branch and version of REST API (and other system-level tooling, which was the same when there was one "core") deployed on the system-under-test, so an appropriate version of test scripts could be picked. If just one of the two repos should deliver such file, it should be "fty-rest" with the revision of the main REST API as the more useful data point. If the file is dropped eventually, it would likely impact some testing scripts, and possibly make field support a bit harder (especially of not the limited amount of end-user GA releases, but of some interim releases and hot-fixes where we'd have more steps to evaluate what is deployed where).

geraldguillaume:
As we now have the green color from Travis. It's time to make a decision.
BTW, in the bellow missing targets, that were not included in this PR, which one do we absolutely need ?
* ci-test-restapi:
*memcheck: test-tntmlm
*check-local: web-link-test check-gitignore run-test-subprocess
* web-test: web-test-deps

What do you think @aquette , @jimklimov, @kkathreen, @bstepankova   ?
We have worked a lot on that topic and we would like to go forward.

jana-rapava:
I think it would be better to fix that SSE bug first, so we can test these changes more thoroughly.

bstepankova:
:+1: for the zproject idea. What about the empty .h files? And what about fty-common? We have been waiting for it for long time. It could make sense extract unnecessary code before any refactoring/improvements so we don't have to touch same code twice.

geraldguillaume:
@jana-rapava we agree that fty-common will support most of the logical structure.
It is more straightforward to do not change now fty-net. 
We recommend to  merge it it first and address the technical debt in a second step.

@jana-rapava SSE bug fix is out of scope of this PR.

geraldguillaume:
@bstepankova  the empty .h file is a consequence of using zproject. As we declared .ecpp servlet as class, zproject generates a related header .h. 
We did it exactly like @thalman did it about etn-ipm1-rest a while ago.
So nothing new.

jimklimov:
> ...is a consequence of using zproject. As we declared .ecpp servlet as class...

As noted in my earlier comment, a better approach is to amend zproject itself to natively support ECPP classes. Because our current solution is an abuse of technology that is not intended to work for this use-case - and if it currently does, it is by a lucky coincidence (and a lot of `Makemodule-local` workarounds) ;)

jimklimov:
> As we now have the green color from Travis. It's time to make a decision.
> BTW, in the bellow missing targets, that were not included in this PR, which one do we absolutely need ?
> ci-test-restapi:

This one seems easy to implement via Makemodule-local, it just calls out a script. Generally, given that potential third-party devs do not currently have access to our CI repo, and since the basic shell-script based testing (in fty-core) still works for the REST API, as far as technology is concerned, better keep this ability... although on second thought, the scripts referred in this recipe are also in that repo (originally fty-core and fty-rest were the same repo), so either should be brought over here or some more complex integration should happen ;)

> memcheck: test-tntmlm

Better add it, to do the unit tests. Even more, better fix it so the tests do not fail as they did before - either identify and fix code bugs (our in fty-rest, or maybe something fishy in 3rd-party projects we rely upon), or determine that these are valgrind false-positives and add exceptions into its config.

> check-local: web-link-test check-gitignore run-test-subprocess

check-gitignore: easy to do via Makemodule-local. Actually, given the direction zproject recently took to aid keeping projects "stylish", and that it currently does a gitignore check "manually" in travis and Jenkins scripts, the better approach is to upstream this snippet and just have it through generated project skeleton. I think I'll handle this direction to generalize the non-fatal style checks in zproject.

web-link-test: this verifies that there are no dangling dependencies in the shared object file, so it would be beneficial to have this. If we pursue upstreaming into zproject the "native" support for ECPP classes (and an ultimate tntnet shared-object deliverable), this sort of test can be part of that upstreamed feature too (may need some magically configured variability WRT usage of e.g. cxxtools or some other third-party dep libraries).

test-subprocess : as long as we have and use the code, its unit-tests should be here. I believe this later morphed and upstreamed into czmq zproc feature, so if our fty-rest code would be converted to use that (and pass the integration tests) then we can afterwards drop this private implementation and its selftests.

> web-test: web-test-deps

These are actively used in FTY dispatcher repository self-tests, so a custom-built webserver (and its configuration and other prerequisites) can be executed and tested by the developer.

bstepankova:
@geraldguillaume I know but the point is just, that we don't have to have the .h file in repo since they are quite many.

aquette:
@bstepankova @jimklimov @jana-rapava @geraldguillaume @barraudl 
* anything that is considered as a future improvement should be logged in an issue, so that we don't forget. This includes zproject improvements to handle ecpp for example
* is there any remaining mandatory mods to this PR prior to approving?
* @kkathreen as for the specific tests (in the present scope, sse for example is part of other PRs), is there anything remaining?
* there are potentially few PRs that did not make its way into here, at least https://github.com/42ity/fty-rest/pull/354. @barraudl could you please check for listing these and cherry-pick or port these too to ensure this PR doesn't generate regression?

thanks to all, imo this is a great improvement!

EldoreiJK:
Tests are fine with the change, merging it to get more results from actual build

PR #352, opened 2018-02-20
Probleme : rack asset creation broken

nicodav:
Solution : 
For a new rack, the sse make a call to fty-metric-cache, but if fty-metric-cache does not known this new asset, it return result = null.
So add a test if result from metric-cache is null or not.

### DIFF
--- a/src/shared/utils_json.cc
+++ b/src/shared/utils_json.cc
@@ -30,14 +30,14 @@
 #include "utils_json.h"
 
 struct Outlet
-  {
-    std::string label;
-    bool label_r;
-    std::string type;
-    bool type_r;
-    std::string group;
-    bool group_r;
-  };
+{
+  std::string label;
+  bool label_r;
+  std::string type;
+  bool type_r;
+  std::string group;
+  bool group_r;
+};
 
 std::string getOutletNumber(const std::string &extAttributeName)
 {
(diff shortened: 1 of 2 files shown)


src/shared/utils_json.cc
```diff
@@ -85,11 +85,17 @@ s_rack_realpower_nominal(
   zstr_free(&uuid);
 
   char *result = zmsg_popstr(msg);
-  if (!streq(result, "OK"))
+  if (NULL == result || !streq(result, "OK"))
   {
     log_warning("Error reply for device '%s', result=%s", name.c_str(), result);
-    zstr_free(&result);
-    zmsg_destroy(&msg);
+    if (NULL != result)
```
EldoreiJK: Please use consistent code style, either check for (var != NULL) or just use (var), not both.

src/shared/utils_json.cc
```diff
@@ -85,11 +85,17 @@ s_rack_realpower_nominal(
...
+    {
+      zstr_free(&result);
+    }
+    if (msg)
```
EldoreiJK: Please use consistent code style, either check for (var != NULL) or just use (var), not both.

jimklimov: Indeed, same style especially in short sequence, is good.

Pedantically speaking, pointers should be compared to NULL explicitly (at least per older standards as they did not require that NULL is numeric zero and some architectures ages ago did have other special values meaning NULL - so `if (ptrname)` did not do what we expect "intuitively" nowadays). Additionally, comparing to NULL adds a bit of type-sanity checking at compile time (avoid using numeric variables where pointers are expected).

PR #367, opened 2018-03-01
Relocate obsoleted packaging metadata to avoid OBS confusion

jimklimov:


EldoreiJK:
We can always delete them later if they are about to be deleted.

PR #386, opened 2018-03-15
Removes duplication of warranty metrics generation.

jana-rapava:
Warranty metric generation was moved to fty-warranty, so the old copy in fty-rest is no longer needed.


src/Makemodule-local.am
```diff
@@ -151,7 +151,7 @@ EXTRA_DIST += $(top_srcdir)/src/web/src/admin_passwd.ecpp
 
 clean-local:
 	for file in $(ECPPCCFILES); do \
-	    /bin/rm -f $(top_builddir)/src/web/src/$$file; \
```
jimklimov: Good catch :)

doc/Makefile.am
```diff
@@ -42,17 +42,17 @@ SUFFIXES=.txt .adoc .xml .xml7 .1 .3 .7
 	asciidoc -d manpage -b docbook -f $(srcdir)/asciidoc.conf \
         -afty-rest_version=@PACKAGE_VERSION@ -o$@ $<
 .xml.1:
-	xmlto --skip-validation man $<
```
jimklimov: Note for others : As discussed with Jana, this flag was my (failed) experiment to let fty-rest build on CentOS, but the issue (of missing DTDs) should probably be addressed from packaging somehow.

Similar problem was recently discussed for CZMQ community, so maybe an upstream solution will soon appear.

PR #401, opened 2018-03-22
Problem: we need to continue after find_access_token.ecpp

barraudl:


jimklimov:
This fixes a bug critical for testing, and is small enough a change to see it should not break in CI. So not worth waiting the lag, merged.

PR #405, opened 2018-03-26
Fix a regression in /alerts/activelist

michal42:
fty-alert-engine may send stale entries in the alert list. While this is
probably a bug in the agent, the fty-rest code is attempting to handle
this. It however fails to advance to the next frame if such entry is
encountered. This fixes commit f1c566744224 ("Rebase").


michal42:
@barraudl, please remember to do two things next time: 1) Do multiple small, logically separated commits, instead of a single jumbo commit and 2) Use the changelog as an opportunity to describe what a given commit is doing. A description worth of 6 letters is not quite adequate when you are changing  1,565 lines in 44 files.

PR #406, opened 2018-03-26
Problem: servlet config is missing

geraldguillaume:


jimklimov:
+1 :)

jimklimov:
Looking further, sorry for this one - I was overly zealous in #387 (we had two different `config.h` files in fact - from autotools and from servlet).

PR #431, opened 2018-04-23
Fix SSE on autodiscovered device deletion

boricj:
Send a SSE notification if an autodiscovered device (i.e. without location) gets deleted.

### DIFF
--- a/src/web/src/sse.cc
+++ b/src/web/src/sse.cc
@@ -298,8 +298,12 @@ std::string Sse::changeFtyProtoAsset2Json(fty_proto_t *asset)
       if(streq(fty_proto_aux_string(asset,"parent","0"),"0"))
       {
         //Check if the asset is a device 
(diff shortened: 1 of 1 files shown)


src/web/src/sse.cc
```diff
@@ -299,7 +299,8 @@ std::string Sse::changeFtyProtoAsset2Json(fty_proto_t *asset)
       {
         //Check if the asset is a device 
         log_debug("Asset type : %s",fty_proto_aux_string(asset,"type","none"));
-        if (streq(fty_proto_aux_string(asset,"type","none"),"device")){
+        if (streq(fty_proto_aux_string(asset,"type","none"),"device") ||
+            streq(fty_proto_aux_string(asset,"type","none"),"none")){
           //Add in the list of asset without location, will send a normal create message
```
michal42: Now there are three calls that retrieve the asset type field. Is it possible to do one call and store it in a variable? Also, the commend still says "Check if the asset is a device" but we are also taking assets with no type field now.

PR #432, opened 2018-04-23
Problem: uuid gets corrupted when inseretd to DB

bstepankova:
Solution: ask fty-asset for it


michal42:
As discussed offline the corruption is most likely caused by using the const char* uuid string after destroying the containing fuuid object. So if this logic is to be moved to fty-asset, please take it into account.

PR #438, opened 2018-05-03
Revert "Problem: We need to access some config with viewer user"

barraudl:
Reverts 42ity/fty-rest#435


jimklimov:
Please try to avoid dumping development branches into upstream project, use personal forks as "origin" to root the PRs.

PR #446, opened 2018-05-22
Problem: there is no licensing limitation support

EldoreiJK:
Solution: implement licensing limitation


src/db/inout/importcsv.cc
```diff
@@ -553,6 +561,65 @@ std::map <std::string, std::string>sanitize_row_ext_names (
...
+    // query values
+    MlmClientPool::Ptr client_ptr = mlm_pool.get ();
+    zmsg_t *request = zmsg_new();
+    zmsg_addstr (request, "LIMITATION_QUERY");
+    zuuid_t *zuuid = zuuid_new ();
+    const char *zuuid_str = zuuid_str_canonical (zuuid);
+    zmsg_addstr (request, zuuid_str);
+    zmsg_addstr (request, "*");
+    zmsg_addstr (request, "*");
+    int rv = client_ptr->sendto ("etn-licensing", "LIMITATION_QUERY", 1000, &request);
```
aquette: this should be somehow configurable, to either be {fty,etn}-licensing

EldoreiJK: Good point. Either it could be via some configuration file, or as a built-in variable. Or we could rename licensing agent so both agents have the same name, which would be the best IMHO - there would be less things to care about.

EldoreiJK: As discussed, we'll keep this for the release and then change it accordingly.

PR #449, opened 2018-05-29
Problem: Need to support HTTP proxy

aquette:
Solution: Add /api/v1/admin/http_proxy with GET/PUT


include/Makefile.am
```diff
@@ -2,7 +2,7 @@
 #  THIS FILE IS 100% GENERATED BY ZPROJECT; DO NOT EDIT EXCEPT EXPERIMENTALLY  #
 #  Read the zproject/README.md for information about making permanent changes. #
 ################################################################################
-nobase_pkginclude_HEADERS = \
```
jimklimov: This should not have changed; is zproject up to date?

PR #450, opened 2018-05-29
[IPMCVG-1883]Problem :  when deleting an asset, we didn't chek any lo…

geraldguillaume:
…gical_asset ref


EldoreiJK:
LGTM, let's see what Jenkins and Travis will think about this PR

PR #453, opened 2018-06-04
Problem: asset name must allow almost all characters

barraudl:

### DIFF
--- a/src/web/src/current.ecpp
+++ b/src/web/src/current.ecpp
@@ -20,6 +20,7 @@
  #><%pre>
 #include <cxxtools/split.h>
 #include <cxxtools/regex.h>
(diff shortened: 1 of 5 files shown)


src/web/src/current.ecpp
```diff
@@ -335,51 +339,45 @@ std::vector<uint32_t> asset_ids;
...
+    serializer.serialize(si).finish();
+    big_json = outJsonStream.str().c_str();
 
-
-    // remove last ","
-    if (!big_json.empty())
-        big_json.pop_back();
 </%cpp>
-{
-   "current":[
```
jana-rapava: I think that if we decide to remove this, we have to take care that UI is updated for compatibility.

PR #454, opened 2018-06-04
Problem: HTTP proxy can also be empty

aquette:
Solution: Don't fail on empty HTTP proxy


jana-rapava:
Merging anyway because:
* the test failure was already there and isn't caused by this changeset
* issue solved by this PR makes deployment really annoying

PR #456, opened 2018-06-05
Problem: We don't wan't to double "escape" the \ and ".

barraudl:


geraldguillaume:
as https://jenkins2.roz.lab.etn.com/blue/organizations/jenkins/42ity-github%2Ffty-rest/detail/PR-456-merge/2/pipeline is green, looks good to me

PR #463, opened 2018-06-12
Allow /api/v1/assets without type or subtype

barraudl:

### DIFF
--- a/src/db/assets/assetr.cc
+++ b/src/db/assets/assetr.cc
@@ -1062,30 +1062,40 @@ int
             "   t.id_type as type_id, "
             "   t.id_subtype as subtype_id "
             " FROM "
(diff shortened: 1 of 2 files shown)


kkathreen:
1) What problem does it solve?
2) Why is noone from Infra team a reviever?
3) Are you aware it will break CI?

geraldguillaume:
Q: What problem does it solve?
A: We are optimizing the performance loading page of asset page which is a blocker for the release. It was a request from UI team. BTW, there is no functional logic to force having type and subtype parameter. Without type or subtype, it means that we will retrieve all assets.
 
Q: Why is noone from Infra team a reviewer?
A: a private response would be provided 

Q : Are you aware it will break CI ?
A: I guess yes because CI should test it. :)

PR #466, opened 2018-06-14
Move default preferences to centralized location

boricj:
This is so that both open-source and proprietary components share the defaults preferences as implemented inside the common API.

### DIFF
--- a/src/db/agentstate/agentstate.cc
+++ b/src/db/agentstate/agentstate.cc
@@ -26,6 +26,8 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 
 namespace persist {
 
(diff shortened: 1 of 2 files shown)


src/db/agentstate/agentstate.cc
```diff
@@ -26,6 +26,8 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 
 namespace persist {
 
+const static std::string DEFAULT_PARAMETERS = "\"preferences\":{\"email\" : \" \", \"telephone\" : \" \", \"organization\" : \" \", \"date\":\"DDMMYYYY\", \"temperature\":\"C\", \"language\":\"en-us\", \"time\":\"24h\"}";
```
jimklimov: Should the empty strings (`email` etc) be in fact not empty `"\0"` but a space char?

boricj: These strings do contains one single space each. While using zero-length strings would be more logical, I simply reused the same exact set of defaults parameters as before.

jimklimov: Ok, thanks for the confirmation. Then at least this should not break any expectations of existing code :) and may be addressed later.

boricj:
After discussing with team, switching to another less hacky approach.

PR #467, opened 2018-06-14
Problem: limitations don't properly sums up amount of power devices

EldoreiJK:
Solution: Add PDU and GENSET to respective SQL queries.


aquette:
trivial change, travis passed & jenkins is lagging => force push

PR #470, opened 2018-06-19

EldoreiJK:

### DIFF
--- a/src/persist/dbhelpers.cc
+++ b/src/persist/dbhelpers.cc
@@ -294,3 +294,41 @@ get_active_power_devices ()
 
     return count;
 }
+
+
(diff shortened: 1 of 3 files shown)


src/persist/dbhelpers.cc
```diff
@@ -294,3 +294,41 @@ get_active_power_devices ()
...
+    }
+    try{
+        tntdb::Statement st = conn.prepareCached(
+            " SELECT "
+            "   v.status,  "
+            " FROM v_bios_asset_element v  "
+            " WHERE v.name=:vname"
+            );
+
+        tntdb::Row row = st.set ("name", element_name).selectRow ();
```
aquette: isn't it "vname" and not "name"?

PR #471, opened 2018-06-19

EldoreiJK:


aquette:
another exception with Jenkins MIA

PR #472, opened 2018-06-19

EldoreiJK:
Backport of 471


aquette:
can't always wait for jenkins

PR #480, opened 2018-06-26
Problem: assets could be created when licensing limitation is hit

EldoreiJK:
Solution: added check for non-existing assets (when creating/importing them)


EldoreiJK:
do not amend to pushed commit :(

PR #482, opened 2018-06-26
Problem: subprocess termination is too ungraceful

jimklimov:
Solution: have hardkill() for original terminate() job, and terminate() children gracefully

Practical use-case: child processes like `start-db-services` currently leave no trace if aborted by timeout, despite all error-handling and traps set up in the script. A `SIGTERM` should at least let the script squeak before it dies.

### DIFF
--- a/src/shared/subprocess.cc
+++ b/src/shared/subprocess.cc
@@ -1,5 +1,5 @@
 /*
-Copyright (C) 2014 Eaton
+Copyright (C) 2014-2018 Eaton
 
 This program is free software; you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


src/shared/subprocess.cc
```diff
@@ -226,9 +215,27 @@ int SubProcess::kill(int signal) {
...
+    // Graceful shutdown
+    auto ret = kill(SIGTERM);
+
+    for (int i = 0; i<20 && isRunning(); i++) {
+        usleep(100);
+        poll(); // update state after a while
+    }
+    if (isRunning()) {
+        // wait is already in hardkill
+        ret = hardkill(SIGKILL);
```
EldoreiJK: hardkill has no arguments, compiled with C++ compiler this throws an error (but not in C compiler)

jimklimov: Good point, and an oversight while renaming the routine and its usage... here's what our CIs are for ;)

Note that these sources use C++ constructs, so would not compile in C anyway.

PR #486, opened 2018-06-28
src/Makemodule-local.am : ensure that bios-passwd gets installed (1.4)

jimklimov:
Solution seems to work for master over the past weeks, but still fails in 1.4 builds sometimes - apply it there too :)


jimklimov:
Jenkins worker timed out :(

PR #488, opened 2018-06-28
 Problem: sysinfo.ecpp sometimes fails dpkg callouts

jimklimov:
* Handle errors more gracefully
* Log more details to help figure out the causes in the OS?
* (TEMPORARY?) Return the error details into JSON output so we can see it in e.g. CI logs, now that the call itself should never fail due to these issues


aquette:
@jimklimov the jenkins failures seems in between a timeout (pr-merge) and some FS (privileges?!) issue (pr-head). Relaunched as a "all-in" to see...

PR #497, opened 2018-07-05
Don't remove source-linked assets

boricj:
This prevents removing a feed when assets are using it.

### DIFF
--- a/src/db/asset_general.cc
+++ b/src/db/asset_general.cc
@@ -615,7 +615,7 @@ db_reply_t
         return reply_delete2;
     }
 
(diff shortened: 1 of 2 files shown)


EldoreiJK:

boricj:
It won't. Foreign key constraints will make the call to `delete_asset_element` _fail_ if there are from-links leftovers (and rollback the asset deletion transaction).

EldoreiJK:
Well, OK, works for me. I've double-checked asset manipulation and there should be no other way to delete asset than this RestAPI call, so this makes sense to keep as is since we don't expect DB to be inconsistent. Thanks.

PR #498, opened 2018-07-05
Don't remove source-linked assets (1.4)

boricj:
This prevents removing a feed when assets are using it.

### DIFF
--- a/src/db/asset_general.cc
+++ b/src/db/asset_general.cc
@@ -615,7 +615,7 @@ db_reply_t
         return reply_delete2;
     }
 
-    auto reply_delete3 = delete_asset_links_all (conn, element_id);
+    auto reply_delete3 = delete_asset_links_to (conn, element_id);
     if ( reply_delete3.status == 0 )
     {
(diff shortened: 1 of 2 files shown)


EldoreiJK:

src/db/asset_general.cc
```diff
@@ -615,7 +615,7 @@ db_reply_t
         return reply_delete2;
     }
 
-    auto reply_delete3 = delete_asset_links_all (conn, element_id);
+    auto reply_delete3 = delete_asset_links_to (conn, element_id);
```
EldoreiJK: This will leave from-links in DB, please fix this.

boricj:
@EldoreiJK Please update your review following #497 discussion.

src/db/asset_general.cc
EldoreiJK: Based on the discussion in master branch PR #497 I agree this should work in the current state of code, and should be enforced with (lib)fty-common-db split.

PR #501, opened 2018-07-06
Prevent deletion of last datacenter in database

boricj:


EldoreiJK:

boricj:
That's an oddly specific function to make, returning the count of datacenters _minus one if the given ID happens to be a datacenter_... That's not something API clients of fty-common-db would need.

It _can_ be something fty-common-db needs internally, but even then I have a hard time finding another use case for this particular SQL query besides helping prevent the deletion of the last datacenter.

EldoreiJK:
Well, I'd consider it to be hidden function. But thinking of it, the whole `delete_dc_room_row_rack` should be moved to common lib to keep consistency should there be need for deleting assets from other parts of the code.

PR #502, opened 2018-07-06
Prevent deletion of last datacenter in database (1.4)

boricj:


src/db/asset_general.cc
```diff
@@ -522,6 +522,25 @@ db_reply_t
     LOG_START;
     tntdb::Transaction trans(conn);
 
+    // Don't allow deleting the last datacenter
+    unsigned numDatacentersAfterDelete = conn.prepareCached(
```
EldoreiJK: Could you please move the count check to subfunction? It's one eligible to be moved to (lib)fty-common-db, so I'd prefer moving functions instead of grabbing blocks from existing ones

PR #504, opened 2018-07-10
Fixing setContentType bug for average

MimiXav:
Signed-off-by: Millieret, Xavier <XavierMillieret@Eaton.com>


MimiXav:
This PR is closed and the fix in on the tntnet's PR

PR #508, opened 2018-07-20
Problem : POST a new rule doesn't work

geraldguillaume:

### DIFF
--- a/src/web/src/alert_rules.ecpp
+++ b/src/web/src/alert_rules.ecpp
@@ -53,13 +53,14 @@ std::string checked_name;
     }
 
     std::string name = request.getArg ("name");
+    if(request.getMethod () == "PUT")
     {
         http_errors_t errors;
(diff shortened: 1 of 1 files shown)


src/web/src/alert_rules.ecpp
```diff
@@ -53,13 +53,14 @@ std::string checked_name;
     }
 
     std::string name = request.getArg ("name");
+    if(request.getMethod () == "PUT")
```
EldoreiJK: Can you please explain a bit why there is no name check when POST is used?

geraldguillaume: This is to be aligned with the RFC-11 specification
It is a common behavior in the REST API to create a new element (likes assets, gpio, ..)

EldoreiJK: Well, I just wanted to make sure there is no hidden hack below that would make problems to POST. If not, LGTM. Polishing might be done later.

PR #511, opened 2018-07-25
Problem: setting time manually ignores the lag

jimklimov:
Solution: correct the value of setting the clock manually by the time it took to process the preceding code in the REST API call processing, with a whole-second granularity.


aquette:
needs #509 to pass

PR #513, opened 2018-07-26
Introduce override flag for last DC deletion check

boricj:
This lets the CI override the last datacenter deletion check through environment variables. Tested by adding `FTY_OVERRIDE_LAST_DC_DELETION_CHECK=y` to `/etc/default/fty` and restarting tntnet.

### DIFF
--- a/src/db/asset_general.cc
+++ b/src/db/asset_general.cc
@@ -30,6 +30,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 
 namespace persist {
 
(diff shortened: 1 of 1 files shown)


boricj:
@jimklimov This should eventually be moved to fty-common-db (along with the rest of fty-rest's direct DB manipulations), but that's for another time.

jimklimov:
Ok, just so we don't forget then :)

kkathreen:
I will merge it, so that the image starts building while we are at meeting :)

PR #516, opened 2018-08-02
Problem: name-id/id-name conversion function were moved to fty-common-db

bstepankova:
Solution: done


bstepankova:
tests will fail because jenkins slaves have not been redeployed yet

PR #518, opened 2018-08-13
Problem: we need to use asset types/subtypes from shared library

bstepankova:
Solution: src/shared/asset_types.cc deleted and fty_common_asset_types used instead


bstepankova:
tests will fail because depends on https://github.com/42ity/fty-common/pull/51

PR #523, opened 2018-08-20
Problem: uptime functions moved to shared library

bstepankova:
Solution:  old functions removed and shared library used instead


bstepankova:
depends on https://github.com/42ity/fty-common-db/pull/16

PR #540, opened 2018-09-12
Fix packaging of shared objects into main package so tntnet finds them

jimklimov:
Part of making our *-rest components packaging follow the same pattern, usable with tntnet in deploy-only packaging.


jimklimov:
The failed build was aborted on Jenkins, re-running now.

PR #543, opened 2018-09-17
Problem: asset_POST: mark strings we need to translate 

bstepankova:
Solution: done

### DIFF
--- a/src/web/src/asset_DELETE.ecpp
+++ b/src/web/src/asset_DELETE.ecpp
@@ -30,6 +30,7 @@
 #include "shared/configure_inform.h"
 #include <fty_common_rest_helpers.h>
 #include <fty_common_db_asset.h>
+#include <fty_common_macros.h>
 </%pre>
 <%thread scope="global">
 asset_manager     asset_mgr;
--- a/src/web/src/asset_DELETE.ecpp
+++ b/src/web/src/asset_DELETE.ecpp
@@ -43,7 +43,7 @@ bool database_ready;
     // verify server is ready
     if (!database_ready) {
         log_debug ("Database is not ready yet.");
-        http_die ("internal-error", "Database is not ready yet, please try again after a while.");
+        http_die ("internal-error", TRANSLATE_ME ("Database is not ready yet, please try again after a while."));
     }
 
     // check user permissions
@@ -57,11 +57,11 @@ bool database_ready;
     {
         std::string id = request.getArg("id");
         if ( id.empty() ) {
-            http_die("request-param-required", "id");
+            http_die("request-param-required", TRANSLATE_ME ("id"));
         }
 
         if ( !persist::is_ok_name (id.c_str ()) )
-            http_die ("request-param-bad", "id", id.c_str (), "valid asset name");
+            http_die ("request-param-bad", TRANSLATE_ME ("id"), id.c_str (), TRANSLATE_ME ("valid asset name"));
         checked_id = id;
     }
     // end sanity checks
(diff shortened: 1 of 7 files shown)


src/web/src/asset_POST.ecpp
```diff
@@ -110,13 +111,13 @@ bool database_ready;
         catch (const std::runtime_error &e) {
             log_error ("%s", e.what());
             std::string msg = "Error during configuration sending of asset change notification. Consult system log.";
-            http_die("internal-error", msg.c_str());
+            http_die ("internal-error", TRANSLATE_ME (msg.c_str()));
```
jana-rapava: I think string should be marked where it's declared, so it would be line 113:
std::string msg = "Error during configuration sending of asset change notification. Consult system log.";

bstepankova: well, I prefer to do it same way as other http_dies to avoid forgetting it when change is needed

EldoreiJK: I'd also say it should be marked at it's declaration, otherwise it would be impossible to get the actual string for translations without running any smart variable assignment in the code.

src/web/src/asset_DELETE.ecpp
```diff
@@ -57,11 +57,11 @@ bool database_ready;
     {
         std::string id = request.getArg("id");
         if ( id.empty() ) {
-            http_die("request-param-required", "id");
+            http_die("request-param-required", TRANSLATE_ME ("id"));
```
bstepankova: are we going to translate ids?

src/web/src/asset_POST.ecpp
EldoreiJK: If we do it simply like this, I can't imagine gathering the strings. We should do it for all the strings at the declaration of such string, or of const strings, etc to be able to "grep" the source for marked strings and then provide them for translation.

src/web/src/asset_DELETE.ecpp
EldoreiJK: Definitely not, otherwise whole system would need to run translations as some parts of the code are using IDs to determine type, to get DB index, etc.

src/web/src/asset_DELETE.ecpp
```diff
@@ -43,7 +43,7 @@ bool database_ready;
     // verify server is ready
```
EldoreiJK: I'd suggest to add #include <fty_common_macros.h> as well as for other files, not to rely on some header, but to be able to track which files are using this

src/web/src/asset_POST.ecpp
```diff
@@ -63,7 +64,7 @@ bool database_ready;
     }
     catch (const std::exception& e) {
         LOG_END_ABNORMAL(e);
-        http_die("bad-request-document", e.what());
+        http_die ("bad-request-document", TRANSLATE_ME (e.what()));
```
EldoreiJK: I'm not sure this might work, as e.what already contains some message we can't get here, As discussed with Jana on IRC, we should use JSON_ME or JSONIFY_ME, or whatever she creates for the purpose (UI might translate the string later, but it will be a problem to get the actual string)

src/web/src/asset_POST.ecpp
```diff
@@ -73,22 +74,22 @@ bool database_ready;
     }
     catch (const std::invalid_argument &e) {
         log_error ("%s", e.what());
-        http_die("bad-request-document", e.what());
+        http_die ("bad-request-document", TRANSLATE_ME (e.what()));
```
EldoreiJK: Same as above

PR #544, opened 2018-09-18
Problem: computed/average: mark strings for translation

bstepankova:
Solution: done

### DIFF
--- a/src/web/src/alert_list.ecpp
+++ b/src/web/src/alert_list.ecpp
@@ -42,7 +42,9 @@
 #include <fty_common_db_dbpath.h>
 #include <fty_common_db_asset.h>
 #include <fty_common_mlm_utils.h>
-#include <fty_common.h>
+#include <fty_common_macros.h>
+#include <fty_common_agents.h>
+#include <fty_common_asset_types.h>
 
 #define RFC_ALERTS_LIST "rfc-alerts-list"
 
@@ -95,7 +97,7 @@ bool database_ready;
     // verify server is ready
     if (!database_ready) {
         log_debug ("Database is not ready yet.");
-        http_die ("internal-error", "Database is not ready yet, please try again after a while.");
+        http_die ("internal-error", TRANSLATE_ME ("Database is not ready yet, please try again after a while."));
     }
 
     // check user permissions
@@ -126,8 +128,8 @@ bool database_ready;
         else
             if (!is_requeststate (state.c_str ())) {
                 log_error ("state = '%s' is not a valid alert state.", state.c_str ());
-                http_die ("request-param-bad", "state", std::string ("value '").append (state).append ("'").c_str (),
-                        std::string ("one of the following values [ ALL | ALL-ACTIVE | ACTIVE | ACK-WIP | ACK-IGNORE | ACK-PAUSE | ACK-SILENCE | RESOLVED ]").c_str ());
+                http_die ("request-param-bad", "state", TRANSLATE_ME (std::string ("value '")).append (state).append ("'").c_str (),
+                          (TRANSLATE_ME (std::string ("one of the following values")).append(std::string(" [ ALL | ALL-ACTIVE | ACTIVE | ACK-WIP | ACK-IGNORE | ACK-PAUSE | ACK-SILENCE | RESOLVED ]"))).c_str ());
             }
 
         // Check 'recursive' parameter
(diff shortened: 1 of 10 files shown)


src/web/src/alert_list.ecpp
```diff
@@ -126,8 +128,8 @@ bool database_ready;
         else
             if (!is_requeststate (state.c_str ())) {
                 log_error ("state = '%s' is not a valid alert state.", state.c_str ());
-                http_die ("request-param-bad", "state", std::string ("value '").append (state).append ("'").c_str (),
-                        std::string ("one of the following values [ ALL | ALL-ACTIVE | ACTIVE | ACK-WIP | ACK-IGNORE | ACK-PAUSE | ACK-SILENCE | RESOLVED ]").c_str ());
+                http_die ("request-param-bad", "state", TRANSLATE_ME (std::string ("value '")).append (state).append ("'").c_str (),
+                          (TRANSLATE_ME (std::string ("one of the following values")).append(std::string(" [ ALL | ALL-ACTIVE | ACTIVE | ACK-WIP | ACK-IGNORE | ACK-PAUSE | ACK-SILENCE | RESOLVED ]"))).c_str ());
```
aquette: are you sure that this construct will stand the pace in the end?

bstepankova: no

src/web/src/alert_list.ecpp
```diff
@@ -126,8 +128,8 @@ bool database_ready;
         else
             if (!is_requeststate (state.c_str ())) {
                 log_error ("state = '%s' is not a valid alert state.", state.c_str ());
-                http_die ("request-param-bad", "state", std::string ("value '").append (state).append ("'").c_str (),
-                        std::string ("one of the following values [ ALL | ALL-ACTIVE | ACTIVE | ACK-WIP | ACK-IGNORE | ACK-PAUSE | ACK-SILENCE | RESOLVED ]").c_str ());
+                http_die ("request-param-bad", "state", std::string (TRANSLATE_ME ("value '")).append (state).append ("'").c_str (),
```
EldoreiJK: I think it would be better to use char *chs = zsys_sprintf("value '%s'", state.c_str());
std::string s(chs); zsys_free(chs); instead as it gives more details about what's going on, instead of current string that translators will see (with the apostrophe without much content given): value '

src/web/src/average.ecpp
```diff
@@ -136,7 +137,7 @@ bool database_ready;
             // possible compinations:
             //      ( relative != "" ) XOR ( start_ts != "" AND end_ts != "" )
             if (csv != "yes") { // for non-csv we parse "relative", for csv by BIOS-3225 we do not yet
-                http_die ("request-param-required", "start_ts or relative");
+                http_die ("request-param-required", zsys_sprintf ((char *) "%s %s %s", "start_ts", TRANSLATE_ME ("or"), "relative"));
```
EldoreiJK: I'd rather use "%s or %s" as start_ts and relative are variables, or is a string content, and "%s or %s" would make more sense to Translator

src/web/src/average.ecpp
```diff
@@ -263,8 +264,8 @@ bool database_ready;
                 if (frame) {
                     if (streq (frame, "BAD_REQUEST")) {
                         zstr_free (&frame);
-                        std::string die = "Data for type = '" + checked_type + "', step = '" + checked_step +
-                        "', source = '" + checked_source + "', element_name = " + " '" + element_name + "'.";
+                        std::string die = TRANSLATE_ME ("Data for type = '") + checked_type + "', step = '" + checked_step +
```
EldoreiJK: This doesn't make much sense to me, as 'type' stands for variable name, but other variable names wouldn't be translated. Also it would be unclear to Translator what the sentence would mean, from: Data for type = '

src/web/src/config.ecpp
```diff
@@ -124,7 +124,7 @@ UserInfo user;
         {
             std::string key = qparam.param ("key");
             if (!key_format.match (key)) {
-                std::string msg = "satisfy format " + key_format_string;
+                std::string msg = TRANSLATE_ME ("satisfy format " + key_format_string);
```
EldoreiJK: This would need to be updated in the end, as getting the strings step (simple grep TRANSLATE_ME) would get either key {"satisfy format "} or {"satisfy format " + key_format_string}, but neither will be valid in runtime, as it will be evaluation of the string, so runtime key wouldn't match grepped key.

src/web/src/average.ecpp
bstepankova: translating start_ts and 'relative' does not make sence

src/web/src/alert_list.ecpp
bstepankova: i don't understand what you mean...

src/web/src/average.ecpp
EldoreiJK: Yes, but translating simply "or" sounds weird to me, without context:
"or" or "$variable1 or $variable2", which sounds better to you if you'd have to translate to Czech?

PR #576, opened 2018-12-10
tntnet.xml : forbid to GET oauth tokens

jimklimov:
Make sure our REST API clients (CI tests, Web-GUI, etc.) only use non-logged POST requests for logging in.


jimklimov:
Note: the old `weblib.sh` in fty-core (that can be used in scripting on a product deployment) was revised to use POST since ages ago.

PR #604, opened 2019-02-26
Implement NUT tracking commands

boricj:


jimklimov:
Sorry, but I do not really see the value of this PR (rather converse - we lose some debuggable/troubleshootable details about failures) nor especially how rearrangement of logging is related to the title "Implement NUT tracking commands" ;-\

aquette:
shared feeling with @jimklimov

boricj:
It is a cleanup that is part of NUT command tracking effort. It just so happens that the original code doesn't need adjustments after implementing command tracking (which was slightly unexpected), so it's limited to cleaning up code and removing HTTP errors 500. I'll add back more logging.

jimklimov:
Ok, thanks for clarification and looking forward to follow-ups :)

aquette:
still, we are losing some log (log_fatal) visibility in the journal.
Note that this could be addressed in the http_die macro

PR #607, opened 2019-03-14
Problem: CSV import may leak memory on error.

jana-rapava:
Solution: Remove the leaks.

### DIFF
--- a/src/db/inout/importcsv.cc
+++ b/src/db/inout/importcsv.cc
@@ -293,6 +293,7 @@ promote_rc0(
     zmsg_addstr (msg, zuuid_str_canonical (uuid));
     if (-1 == client->sendto ("fty-info", "info", 1000, &msg)) {
         zuuid_destroy (&uuid);
+        zmsg_destroy (&msg);
         throw std::runtime_error(TRANSLATE_ME ("Sending INFO message to fty-info failed"));
     }
     touch_fn(); // renew request watchdog timer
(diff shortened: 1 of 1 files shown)


src/db/inout/importcsv.cc
```diff
@@ -316,13 +319,20 @@ promote_rc0(
...
+        zmsg_destroy (&resp);
         throw std::runtime_error(TRANSLATE_ME ("Response on INFO message from fty-info miss zhash frame"));
     }
     zhash_t *info = zhash_unpack(frame_infos); // serial, hostname, ip.[1-3]
-    char * item = (char *)zhash_first(info);
+    /*char * item = (char *)zhash_first(info);
     while (NULL != item) {
         item = (char *)zhash_next(info);
-    }
+    }*/
```
jimklimov: Question: I am not sure what this code was supposed to do (read all items from the hash, to let it be freed later, or something?) and how it's commenting-away aligns with the rest of this commit which coherently frees potentially allocated blocks of memory.

jimklimov: So from discussion it seems like partial survivor of an earlier development-debugging aid to count or print out the hash contents (without the actual print) or trace them easily in a debugger, at least this clause came with a commit fixing unit tests.

There do not seem to be any requirements about "touching" all data in a hash (e.g. so ZMQ housekeeping would allow its auto-freeing) that was my initial suspicion about why this loop is there.

So this comment-away can stay :) and we'll keep an eye on whether something regresses ... which is hopefully not likely, and then this is just a pure small performance improvement :)

PR #608, opened 2019-03-15
Problem : fix memory leak - stage 2

geraldguillaume:
Problem : zmsg_popstr is not awlays freed
Solution : use MlmGuard approach 

Problem : zpoller is not awlays freed
Solution : use MlmGuard approach 

Problem: mlm_client_recvx should free its returned value
solution : fix it


geraldguillaume:
The PR is not really clean, so redo it

PR #610, opened 2019-03-18
Fix memory leak in admin/myuser.

jana-rapava:


src/web/src/my_profile.ecpp
```diff
@@ -151,6 +151,7 @@
                             std::string msg = TRANSLATE_ME ("Cannot save config file ") + it.first;
                             http_die ("internal-error", msg.c_str());
                         }
+                        zconfig_destroy (&config);
```
geraldguillaume: @jana-rapava could  you please cross check with config.ecpp.
As you added 'zconfig_destroy' here, it could make sense to do it also here : https://github.com/42ity/fty-rest/blob/master/src/web/src/config.ecpp#L206

There is also another difference, 'utils::config::roots_destroy (roots);' is always called in config.eccp (https://github.com/42ity/fty-rest/blob/master/src/web/src/config.ecpp#L211), it could make  sense do call around https://github.com/42ity/fty-rest/blob/master/src/web/src/my_profile.ecpp#L156
What do you think ?

PR #612, opened 2019-03-19
Problem: Data string is incorrectly terminated.

jana-rapava:
Solution: Terminate it correctly.

### DIFF
--- a/src/db/agentstate/agentstate.cc
+++ b/src/db/agentstate/agentstate.cc
@@ -68,58 +68,6 @@ static int
     }
 }
 
-static int
-    select_agent_info(
-        tntdb::Connection &conn,
-        const std::string &agent_name,
-        void             **data_ptr,
-        size_t            &size
-        )
-{
-    LOG_START;
-
-    if ( data_ptr == NULL )
-        return 5;
-    try{
-        *data_ptr = NULL;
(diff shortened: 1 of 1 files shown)


boricj:
Shouldn't `strdup`'ing the original string just work, assuming it is zero-terminated?

jana-rapava:
@boricj from my investigations, it's not zero terminated - it's just some binary blob stored in the DB.

boricj:
I don't see why the original code wouldn't work correctly nor how the fix is functionally different except for adding one more null terminator... Is this really the underlying cause of the data corruption?

Then again, my brain might not be properly working at the end of the day >_<

jana-rapava:
Code from yesterday does not work and is not a good point for comparison.
Original code, which worked but had a memory leak, is here: https://github.com/42ity/fty-rest/commit/01db70aa46272e075ffc642f0a8682744fb9f0e1 .
Refactoring won't get us rid of the problem that we have to turn binary blob into nice and valid std::string - I don't know a way of doing it without null terminators.

boricj:
Cast the pointer to `const char*` and pass the size as the second parameter of the `std::string` constructor: https://en.cppreference.com/w/cpp/string/basic_string/basic_string

PR #615, opened 2019-03-20
Problem: We shouldn't mix new and free ().

jana-rapava:
Solution: Use delete [].

### DIFF
--- a/src/web/src/auth_verify.ecpp
+++ b/src/web/src/auth_verify.ecpp
@@ -68,7 +68,7 @@ if(!checked_access_token.empty()) {
         user.uid (uid);
         user.gid (gid);
         user.login (user_name);
-        free (user_name); user_name = NULL;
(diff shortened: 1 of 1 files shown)


boricj:
I approve the fix itself, but I highly disapprove of `tokens::verify_token` in the fty-common-rest repository. Why does that method (which *is* C++ code) take a `char **user_name` parameter that is callee-set to a `new[]`-allocated buffer? Why does it tries to call `free()` on that buffer in case of error? Why does that method just doesn't take a `std::string*` and sidestep all that insanity instead?

I'm chasing down some elusive memory corruption inside `tntnet` that suddenly decided to manifest today in our containers and makes the UI team sad. What I see inside `tokens::verify_token` makes me both sad and weary of what other code quality issues are waiting in the shadows to bite us in the future...

PR #617, opened 2019-04-23
Provide proper error when asset name is too long.

jana-rapava:


src/db/inout/importcsv.cc
```diff
@@ -725,8 +725,13 @@ static std::pair<db_a_elmnt_t, persist::asset_operation>
 
     auto ename = cm.get(row_i, "name");
     if (ename.empty ()) {
-        std::string received = TRANSLATE_ME ("<empty>");
-        std::string expected = TRANSLATE_ME ("<unique, non empty value>");
+        std::string received = TRANSLATE_ME ("empty value");
+        std::string expected = TRANSLATE_ME ("unique, non empty value");
+        bios_throw("request-param-bad", "name", received.c_str (), expected.c_str ());
+    }
+    if (ename.length() > 50) {
```
perrettecl: else if ?

perrettecl: not necessary.

PR #621, opened 2019-05-15
[S/S] Problem: RC0 promotion results in RC0 being stolen by imported DC

EldoreiJK:
Solution: Disabled RC0 promotion for CSV import

### DIFF
--- a/src/db/inout/importcsv.cc
+++ b/src/db/inout/importcsv.cc
@@ -1468,7 +1468,10 @@ void
 
     // if there are rackcontroller in the import, promote one of it to rackcontroller-0 if not done already
     MlmClient client;
-    int rc0 = promote_rc0(&client, cm, touch_fn);
+    // rc0 promotion disabled due to multiple datacenter support
+    // - hard to decide whether it's import of DC after crash or another DC, RC location must be set manually
+    int rc0 = -1;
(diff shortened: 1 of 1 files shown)


src/db/inout/importcsv.cc
```diff
@@ -1468,7 +1468,10 @@ void
 
     // if there are rackcontroller in the import, promote one of it to rackcontroller-0 if not done already
     MlmClient client;
-    int rc0 = promote_rc0(&client, cm, touch_fn);
+    // rc0 promotion disabled due to multiple datacenter support
+    // - hard to decide whether it's import of DC after crash or another DC, RC location must be set manually
+    int rc0 = -1;
+    //int rc0 = promote_rc0(&client, cm, touch_fn);
```
jimklimov: Suggestion: Can it be checked if an rc-0 already exists and has a parent (DC/rack/...)? Then if it only exists but has no parent (empty database state), promote it according to imported data (especially if we can identify one of imported RCs that has a parent in CSV as the current RC); otherwise keep existing links.

EldoreiJK: @jimklimov possibly, but that needs to be discussed whether we want to keep smart matching of RC or not

EldoreiJK: @jimklimov the decision is to remove smart matching of RC for now

PR #627, opened 2019-06-14
Add counter for active asset

perrettecl:


packaging/debian/control
```diff
@@ -118,8 +118,8 @@ Depends:
 Description: fty-rest debugging symbols
  This package contains the debugging symbols for fty-rest:
  common core rest api for 42ity project.
-
-# Customization over zproject-generated code follows:
+ 
+ # Customization over zproject-generated code follows:
```
jimklimov: Leading whitespace appeared, may be toxic for debian parser. Better fix this bit back.

aquette: Not really toxic but having a special handling as sub section / text

project.xml
```diff
@@ -50,7 +54,7 @@
         <option name = "use_checkout_explicit" value = "1" />
     </target>
 
-    <classfilename keep-tree = "true" pkgincludedir = "true" use-cxx = "true" pretty-print = "no" source-extension = "cc" />
+    <classfilename use-cxx-gcc-4-9 = "true" keep-tree = "true" pkgincludedir = "true" use-cxx = "true" pretty-print = "no" source-extension = "cc" />
```
jimklimov: I don't see uses of `std::regex` in this codebase; do we require (or soon plan to require) gcc-4.9 for `fty-rest`? So far it was ok with default 4.8 I believe.

PR #638, opened 2019-06-26
Fix uuid issue alert_templates

perrettecl:
Signed-off-by: Clement Perrette <clementperrette@eaton.com>


perrettecl:
We do not need to remove uuid, as it's remove by receiving function

PR #646, opened 2019-07-04
Problem: DC asset export file contains DC ID instead of name in the f…

jana-rapava:
…ilename.

Solution: Fix it.

### DIFF
--- a/src/web/src/asset_export.ecpp
+++ b/src/web/src/asset_export.ecpp
@@ -24,10 +24,12 @@
  */
  #><%pre>
 #include <string>
+#include <regex>
 #include <string.h>
 #include <time.h>
 #include <fty_common_rest_helpers.h>
 #include <fty_common_macros.h>
+#include <fty_common_db_asset.h>
 #include "src/db/inout.h"
 #include "shared/utils.h"
 </%pre>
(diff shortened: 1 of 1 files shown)


src/web/src/asset_export.ecpp
```diff
@@ -81,10 +82,14 @@ bool database_ready;
         *c = '-';
     }
 
-    if (!dc.empty())
-        reply.setHeader(tnt::httpheader::contentDisposition, "attachment; filename=\"asset_export_" + dc + "_" + std::string{buf} + ".csv\"");
-    else
+    if (!dc.empty()) {
+        std::string dc_ename;
+        DBAssets::name_to_extname (dc, dc_ename);
```
perrettecl: Please ensure that we do not have spaces or tab in the name:
```
  //escape special caracteres
  dc_ename = std::regex_replace(dc_ename, std::regex("( |\t)"), "_");
```

src/web/src/asset_export.ecpp
```diff
@@ -81,10 +83,16 @@ bool database_ready;
...
+        //escape special characters
+        dc_ename = std::regex_replace(dc_ename, std::regex("( |\t)"), "_");
```
jimklimov: So now this component officially requires at least gcc-4.9 :)

PR #647, opened 2019-07-04
Display only active network interfaces.

jana-rapava:

### DIFF
--- a/src/web/src/netcfg.ecpp
+++ b/src/web/src/netcfg.ecpp
@@ -96,7 +96,8 @@ if(rex_ncfg.match(request.getUrl())) {
             out += "\"" + iface + "\",";
         }
     }
-    out [out.size() - 1] = ' ';
+    // remove trailing comma
+    out.pop_back ();
 </%cpp>
 { "netcfgs": [ <$$ out $> ] }
 <%cpp>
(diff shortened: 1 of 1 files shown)


src/web/src/netcfg.ecpp
```diff
@@ -74,20 +75,30 @@ if(!nil.empty()) {
...
+    std::vector<std::string> all_interfaces;
+    cxxtools::split(",", all_interfaces_str, std::back_inserter(all_interfaces));
+
+    std::string out;
+    for (std::string iface : all_interfaces) {
+        if (active_interfaces.find (iface) != active_interfaces.end ()) {
+            out += "\"" + iface + "\",";
+        }
+    }
+    out [out.size() - 1] = ' ';
```
jeffdegott: what if out.size() = 0?
In fact, this is to erase the comma added previously. Code at line 96 shall be:.
if (!out.empty()) { out += ","; }
out += "\"" + iface + "\""; 
I am wrong?

jana-rapava: Your point is valid but cosmetical, since we are never going to have a system without any network interfaces.

jeffdegott: Depends on the consistency of the active_interfaces string. What if get_ifaces() is buggy, returning an empty string for example... My opinion is to support the worst of cases.
It's up to you  ;-)

PR #648, opened 2019-07-09
Problem: alert/activelist returns only array of commas

EldoreiJK:
Solution: ixed list full of commas issue, visual fixes

### DIFF
--- a/src/web/src/alert_list.ecpp
+++ b/src/web/src/alert_list.ecpp
@@ -121,16 +121,18 @@ s_response_list (zmsg_t *resp, tntdb::Connection& connection) {
            continue;
         }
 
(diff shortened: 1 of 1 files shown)


src/web/src/alert_list.ecpp
```diff
@@ -121,16 +121,18 @@ s_response_list (zmsg_t *resp, tntdb::Connection& connection) {
...
+        std::string jsonAlert = getJsonAlert (connection,decoded);
 
-        if (jsonAlert.empty()){
+        if (jsonAlert.empty ()){
            fty_proto_destroy (&decoded);
            frame = zmsg_pop (resp);
            continue;
         }
+
+        json += jsonAlert;
```
jana-rapava: good catch!

PR #649, opened 2019-07-10
Backport master to IPM-2.0.0

perrettecl:


src/web/src/sse.cc
```diff
@@ -181,7 +181,7 @@ std::string Sse::loadAssetFromDatacenter()
     }
     log_debug("=== end ===");
   }
-  //Get all devices without location
```
aquette: WARNING: any modification to SSE here MUST also be applied to etn-ipm2-rest similar files.
Otherwise, this will be overriden by that last in the resulting images!
However, the below resemble to a backport of another PR. @jeffdegott plz check and report, thx.

jeffdegott: I confirm. etn-ipm2-rest SSE code (etn_sse.ecpp, sse.cc)  *must* be synchronized with fty-rest equivalent.

perrettecl: @jeffdegott, @aquette The modification from this file are already on etn-ipm2-rest. I also discover that HOTFIX_PQSWMBT_3166 is on etn-ipm2-rest but not on fty-rest

jeffdegott: Exact. The hotfix should be backported to the fty repo after release.

PR #658, opened 2019-07-16
HOTFIX which change the ename inside translation section of alert rules before send it to FE

perrettecl:


src/web/src/alert_rules_list.ecpp
```diff
@@ -196,71 +196,104 @@ if (streq (part, "LIST")) {
...
-            
-            cxxtools::SerializationInfo siTmp;
-            deserializerTmp.deserialize(siTmp);
-
-            std::stringstream output;
-            cxxtools::JsonSerializer serializerTmp(output);
-            serializerTmp.beautify(false);
-            serializerTmp.serialize(siTmp).finish();
+            std::string data(part.get());
+            //log_debug ("Part: %s \n\n", data.c_str());
```
aquette: could switch to log_trace() if it's really verbose and too much for debug, though still possibly useful!

PR #661, opened 2019-07-22
Problem: RestAPI calls are not auditable

EldoreiJK:
Solution: add audit logging calls to show what user did (or didn't do)
Note: for review ignore whitespace changes

### DIFF
--- a/src/web/src/asset_DELETE.ecpp
+++ b/src/web/src/asset_DELETE.ecpp
@@ -31,6 +31,7 @@
 #include <fty_common_rest_helpers.h>
 #include <fty_common_db_asset.h>
 #include <fty_common_macros.h>
+#include <fty_common_rest_audit_log.h>
 </%pre>
 <%thread scope="global">
 asset_manager     asset_mgr;
@@ -56,14 +57,16 @@ bool database_ready;
     std::string checked_id; // user-friendly identifier
 
     // sanity check
+    std::string id = request.getArg ("id");
     {
-        std::string id = request.getArg("id");
-        if ( id.empty() ) {
-            http_die("request-param-required", "id");
+        if ( id.empty () ) {
+            log_info_audit ("Request DELETE asset FAILED");
+            http_die ("request-param-required", "id");
         }
 
         if ( !persist::is_ok_name (id.c_str ()) ) {
             std::string err = TRANSLATE_ME ("valid asset name");
+            log_info_audit ("Request DELETE asset id %s FAILED", id.c_str ());
             http_die ("request-param-bad", "id", id.c_str (), err.c_str ());
         }
         checked_id = id;
@@ -73,10 +76,12 @@ bool database_ready;
     // get id by name
     int64_t dbid = DBAssets::name_to_asset_id (checked_id);
     if (dbid == -1) {
+        log_info_audit ("Request DELETE asset id %s FAILED", id.c_str ());
         http_die ("element-not-found", checked_id.c_str ());
     }
     if (dbid == -2) {
         std::string err =  TRANSLATE_ME ("Database failure");
+        log_info_audit ("Request DELETE asset id %s FAILED", id.c_str ());
         http_die ("internal-error", err.c_str ());
     }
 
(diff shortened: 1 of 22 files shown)


src/web/src/asset_actions_POST.ecpp
```diff
@@ -100,59 +107,64 @@ bool database_ready;
...
-    ZmsgGuard msg(mlm_client_recv(client));
-    ZstrGuard status(zmsg_popstr(msg));
-    ZstrGuard uuid(zmsg_popstr(msg));
-    if (!streq(status, "OK")) {
-        ZstrGuard errMsg(zmsg_popstr(msg));
-        std::string err = std::string("Request to fty-nut-command failed: ") + errMsg.get() + ".";
+    ZmsgGuard msg (mlm_client_recv (client));
+    ZstrGuard status (zmsg_popstr (msg));
+    ZstrGuard uuid (zmsg_popstr (msg));
+    if (!streq (status, "OK")) {
```
jeffdegott: Not sure that streq() support NULL arg. + ZstrGuard() usage. Be carefull here.

jeffdegott:
We use log_info_audit()  to log both successes and failures
Is there some log_error_audit() to log failures?

EldoreiJK:
For sure we can use log_audit_error, I just don't know whether that's our aim as this is different log level.

src/web/src/asset_actions_POST.ecpp
EldoreiJK: I don't know as well, but that's what's been used in the system, this is no new code.. As per first comment, ignore whitespace changes to see added code. But for sure we can discuss fixing things in rest here, but it's not aim of this PR

jimklimov: Note that effectively Jeff's question goes for implementation of `strcmp()` because in ZMQ stack we `#define streq(s1,s2)    (!strcmp ((s1), (s2)))`. The standard does not define `strcmp` behavior for `NULL` inputs, so anything is allowed to happen, from a sanity check to a segmentation fault, any answer is correct.

jimklimov: Looking at czmq code at https://github.com/42ity/czmq/blob/v3.0.2-FTY-master/src/zmsg.c#L407 and https://github.com/42ity/czmq/blob/v3.0.2-FTY-master/src/zframe.c#L238 there are several scenarios that `zmsg_popstr()` would return `NULL`, and we do not check nor at least assert for such cases anywhere, so Jeff's concern is valid :)

jimklimov: All that said, if stars align and reading (dereferencing) data from NULL pointer reads some garbage (or the OS-mapped zero-filled page for example), so the program does not crash on this read, the `strcmp()` would quickly and highly-probably find there is no `OK` in the `status` so decide to ... (whatever it needs the clause for).

src/web/src/asset_DELETE.ecpp
```diff
@@ -56,14 +57,16 @@ bool database_ready;
     std::string checked_id; // user-friendly identifier
 
     // sanity check
+    std::string id = request.getArg ("id");
     {
```
jimklimov: Why this brace (and scoping) move? I see we log `id` below, but could actually log the `checked_id` instead as it is the value we use and trust (or bail out in this clause). The idea for `checked*` stuff was that the original untrusted input variables do not leak around (checked ones might be sanitized, etc.).

jimklimov: Also it seems that `id` is logged in many other files, where this apparently does not cause scoping issues... maybe it should have ;)

EldoreiJK: Scope is still local file, just a bit extended to be able to log it. While I agree we can use this variable in the block, and then use checked one, the result should still be the same, right?

src/web/src/asset_actions_POST.ecpp
EldoreiJK: While the question is right, this is a wrong place for such a question. I think this should be promoted to a task to check RestAPI for such uses of possibly invalid recvd data

jimklimov: Sounds correct.

src/web/src/asset_DELETE.ecpp
jimklimov: Should be... but only as long as the `checked_*` values we use are a guaranteed copy, not a conversion product, of the input `*` query arguments.

EldoreiJK:
merging to test, will revert in case of issues

src/web/src/asset_PUT.ecpp
```diff
@@ -54,9 +55,10 @@ bool database_ready;
             {BiosProfile::Admin,     "U"}
             };
     CHECK_USER_PERMISSIONS_OR_DIE (PERMISSIONS);
-    std::string id = request.getArg("id");
+    std::string id = request.getArg ("id");
 
     if (id.empty () || !persist::is_ok_name (id.c_str ()))
+        log_info_audit ("Request PUT asset FAILED");
```
jimklimov: Oopsie! No parenthesis, no luck!

PR #669, opened 2019-08-27
Problem: for audit, systemctl is not "CREATE"d with a POST

jimklimov:
Solution: ... but rather "MANAGE"d -- redefine the string for log to make sense


jimklimov:
good point, thanks

PR #676, opened 2019-11-13
Make PUT netcfg return updated DNS settings.

jana-rapava:
When updating DNS settings, they are put into dns-nameservers field of /etc/network/interfaces.
Then after network restart is finished, resolvconf picks them up, merges them with data from DHCP and puts everything into /etc/resolv.conf.

Before this change, both GET and PUT were reading DNS settings from /etc/resolv.conf. Because the network restart is asynchronous to PUT REST API call,
this meant that PUT might have return old DNS settings. Now we are returning the new settings even though they are not used before restart - that is
consistent with how we e.g. return new static IP address even though we are using old one until restart.

### DIFF
--- a/src/web/src/netcfg.ecpp
+++ b/src/web/src/netcfg.ecpp
@@ -314,7 +314,29 @@ AUG_GET ("method", method);
 AUG_GET ("address", ip);
 AUG_GET ("netmask", netmask);
 AUG_GET ("gateway", gateway);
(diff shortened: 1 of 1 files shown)


jana-rapava:
@jimklimov : 0 triggers a different codepath, 1 and more look fine based on tests

jimklimov:
Thanks for confirmation :)

PR #684, opened 2019-11-21
Problem: UPS outlets serialization is missing

aquette:
Solution: Fix it


aquette:
forcing merge, since 1 Jenkins is green, travis is green and the other jenkins is just an unsync'ed system

PR #691, opened 2020-02-11
Clean config REST endpoint

boricj:
This is a refactoring of the /api/v1/admin/config REST endpoint, allowing the API to store arbitrary JSON objects and removing most hard-coded logic from the endpoint.

Due to unfortunate timing issues, there are a couple of points to address as soon as possible:
* lobby upstream czmq for marking zconfig_remove() as stable
* update 42ity's czmq with upstream's czmq
* remove local zconfig_remove() implementation hack
* split mapping data into separate files inside file-system
* prune dead code

This modification should be backwards-compatible.

### DIFF
--- a/src/web/src/config.ecpp
+++ b/src/web/src/config.ecpp
@@ -38,64 +38,378 @@
 #include <fty_common_rest_utils_web.h>
 #include <fty_common_rest_audit_log.h>
 
-// define json serialization objects
-struct Array
-{
-    struct Config
-    {
-        std::string key;
-        std::vector <std::string> value;
-    };
-    Config config;
+/**
+ * FIXME: Move that file into the file-system and maybe split it per components.
+ *
+ * fty-rest doesn't seem to have any way to simply drop files into the
+ * file-system, keep this in memory until that can be reworked.
+ */
+static const char* mapping = R"xxx(
+# FTY
+
+# compatibility
+BIOS_SNMP_COMMUNITY_NAME = ""
+    filepath = "/etc/fty-email/fty-email.cfg"
+    treepath = "snmp/community"
+language = ""
+    filepath = "/etc/fty-email/fty-email.cfg"
+    treepath = "server/language"
+
+# nut
+BIOS_NUT_POLLING_INTERVAL = ""
+    filepath = "/etc/fty-nut/fty-nut.cfg"
+    treepath = "nut/polling_interval"
+
+# agent-smtp
+BIOS_SMTP_SERVER = ""
+    filepath = "/etc/fty-email/fty-email.cfg"
+    treepath = "smtp/server"
+BIOS_SMTP_PORT = ""
+    filepath = "/etc/fty-email/fty-email.cfg"
+    treepath = "smtp/port"
+BIOS_SMTP_ENCRYPT = ""
+    filepath = "/etc/fty-email/fty-email.cfg"
+    treepath = "smtp/encryption"
+BIOS_SMTP_VERIFY_CA = ""
+    filepath = "/etc/fty-email/fty-email.cfg"
+    treepath = "smtp/verify_ca"
+BIOS_SMTP_USER = ""
+    filepath = "/etc/fty-email/fty-email.cfg"
+    treepath = "smtp/user"
+BIOS_SMTP_PASSWD = ""
+    filepath = "/etc/fty-email/fty-email.cfg"
+    treepath = "smtp/password"
+BIOS_SMTP_FROM = ""
+    filepath = "/etc/fty-email/fty-email.cfg"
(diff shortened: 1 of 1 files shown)


src/web/src/config.ecpp
```diff
@@ -38,64 +38,378 @@
...
+
+# fty-session
+FTY_SESSION_TIMEOUT_NO_ACTIVITY = ""
+    filepath = "/etc/fty/fty-session.cfg"
+    treepath = "timeout/no_activity"
+FTY_SESSION_TIMEOUT_LEASE = ""
+    filepath = "/etc/fty/fty-session.cfg"
+    treepath = "timeout/lease_time"
+
+# IPM2
```
MimiXav: IPM2? inside fty ?

boricj: Ideally this data would be split across files in a directory where each component declares its mapping, but now's not the time to do large-scale refactoring across multiple repositories (this needs to be backported to the 2.1.0 release).

src/web/src/config.ecpp
```diff
@@ -38,64 +38,378 @@
...
-{
-    struct Config
-    {
-        std::string key;
-        std::vector <std::string> value;
-    };
-    Config config;
-};
+/**
+ * XXX: Move that file into the file-system and maybe split it per components.
```
aquette: prefer some TODO / FIXME tag instead of XXX

src/web/src/config.ecpp
```diff
@@ -38,64 +38,378 @@
...
-    {
-        std::string key;
-        std::string value;
-    };
-    Config config;
+# compatibility
+BIOS_SNMP_COMMUNITY_NAME = ""
+    filepath = "/etc/fty-email/fty-email.cfg"
+    treepath = "snmp/community"
+language = ""
```
aquette: indent issue

src/web/src/config.ecpp
```diff
@@ -38,64 +38,378 @@
...
-    };
-    Config config;
-};
+/**
+ * XXX: Move that file into the file-system and maybe split it per components.
+ *
+ * fty-rest doesn't seem to have any way to simply drop files into the
+ * file-system, keep this in memory until that can be reworked.
+ */
+static const char* mapping = R"xxx(
```
aquette: all these should go into some external descriptor file a la dotD, so that we can ship some FTY version, and complete with some IPM2 drops (coming from an etn pkg)

src/web/src/config.ecpp
```diff
@@ -38,64 +38,378 @@
...
+    treepath = "policy/task_timeout"
+ETN_AUTOMATION_NOTIF_EMAIL = ""
+    filepath = "/etc/fty/etn-automation.cfg"
+    treepath = "notifications"
+)xxx";
+
+// JSON <-> ZPL helpers
+
+/**
+ * XXX: pretend we have zconfig_remove().
```
aquette: ditto TODO/FIXME

src/web/src/config.ecpp
```diff
@@ -38,64 +38,378 @@
...
+            zconfig_set_value(root, nullptr);
+            break;
+
+        case cxxtools::SerializationInfo::Value:
+            si.getValue(value);
+            zconfig_set_value(root, "%s", value.c_str());
+            break;
+
+        case cxxtools::SerializationInfo::Object:
+            // XXX: Add a '= ""' to the entry, should correct augeas lens.
```
aquette: ditto TODO/FIXME

src/web/src/config.ecpp
```diff
@@ -38,64 +38,378 @@
...
+            zconfig_set_value(root, "%s", "");
+
+            for (const auto& childSi : si) {
+                zconfig_new(childSi.name().c_str(), root) <<= childSi;
+            }
+            break;
+
+        case cxxtools::SerializationInfo::Array:
+            // XXX: Add a '= ""' to the entry, should correct augeas lens.
```
aquette: ditto TODO/FIXME

src/web/src/config.ecpp
```diff
@@ -38,64 +38,378 @@
...
+
+    zconfig_t* root = zconfig_load(it->second.first.c_str());
+    if (!root) {
+        bios_throw("element-not-found", it->second.first.c_str());
+    }
+
+    zconfig_t* child = zconfig_locate(root, it->second.second.c_str());
+    if (!child) {
+        /**
+         * XXX:Entry is missing, force its creation.
```
aquette: ditto TODO/FIXME

src/web/src/config.ecpp
boricj: Actually not, it's really a legacy configuration mapping called `language`.

PR #694, opened 2020-02-13
[DNMY] Fix config endpoint refactoring (2.1.0)

boricj:
Backport of #693.


aquette:
transiently disabled Travis REQ since it was taking too long on allowed failures, and all the rest was green

PR #695, opened 2020-02-26
Fix error handling of config endpoint

boricj:
Note: this is a modest clean-up of the config endpoint to get the error handling semantics right. The rest of the clean-up is still a post-release task.

### DIFF
--- a/src/web/src/config.ecpp
+++ b/src/web/src/config.ecpp
@@ -438,40 +438,40 @@ UserInfo user;
     ConfigMap mapping = loadMapping(mappingZconfig);
     zconfig_destroy(&mappingZconfig);
 
-    ///////////////////
-    ///     GET     ///
-    ///////////////////
-    if (request.isMethodGET ()) {
-        std::string key = qparam.param ("key");
-        if (key.empty()) {
-            http_die ("request-param-required", "key");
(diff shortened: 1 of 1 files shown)


src/web/src/config.ecpp
```diff
@@ -484,14 +484,18 @@ UserInfo user;
...
         }
-        catch (const std::exception &e) {
+        http_die_idx (e.idx, e.what ());
+    }
+    catch (const cxxtools::SerializationError &e) {
+        if (request.isMethodPOST ()) {
             log_error_audit ("Request CREATE config FAILED");
-            http_die ("bad-request-document", e.what ());
         }
+        http_die ("bad-request-document", e.what ());
```
jimklimov: If we throw the request-document message for non-POST requests, would it be correct?
Otherwise LGTM, the change is small and meaningful :)

boricj: Since we don't deserialize the body on a GET request, we shouldn't be able to land here with a GET. Still, can't hurt to have it at this level.

jimklimov:
Travis is green WRT non-shadow stack of tests. Jenkins in progress, no complaints yet. Looks good to fast-merge.

PR #698, opened 2020-03-10
Fix importcsv message

perrettecl:


aquette:
@perrettecl thanks for merging, but please let me do so. This was misleading for validation this morning (b46 RC2 Vs b47) ;)

PR #713, opened 2020-05-14
getlog_GET.ecpp : extend exec_journalctl_command() …

jimklimov:
…for various data-export types;
+ define a "journald-all" log which is more detailed and not truncated by time

TODO: Verify it does not explode ;) Perhaps test in CI too.

Note that the data available in "journald-all" is presumed to be subject to systemd log expiration (so after some hours or days of uptime, not all might be there anymore - more so on devel systems with DEBUG messages enabled), and the export to temporary text file for further serving (and maybe compression) can take some time and a lot of space and I/O. There is potential to bring system down by hogging up all tmpfs. Further extensions in this area might be to only allow this data served as a `journald-all.gz` (not as plaintext) and as part of `ALL` archive, and so we can compress it during export, to reduce the storage impact.

### DIFF
--- a/src/web/src/getlog_GET.ecpp
+++ b/src/web/src/getlog_GET.ecpp
@@ -287,15 +287,20 @@ class _clonedcode_Compressor
 }; // class _clonedcode_Compressor
 
 // journald logs export to logfile
-int exec_journalctl_command (const std::string& logfile)
+int exec_journalctl_command (const std::string& logfile, bool truncate = true, bool extensive = false)
 {
     // 'www-data' user must be part of the 'systemd-journal' group, to let journalctl access journald files (see fty-core)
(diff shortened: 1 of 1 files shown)


boricj:
Have you checked for potential cyber-security issues? This is dumping loads of logs into `/tmp`, are they safe from prying eyes?

jimklimov:
Just as safe as other logs we dumped there before. I assume, not very safe if someone has a login/scp/... access to the system. All known reasonable security doctrines assume that if your system is networked, or accessible physically, it is hacked anyway.

That said, it may make sense to do the equivalent of `touch+chmod` on the generated files, so only the creating (web-server) user can access them, before flooding the data into there. Probably impacts both `journald-` and `ALL` exports so can be done by another PR.

PR #715, opened 2020-05-19
Fix listening interface, set to localhost

aquette:
Signed-off-by: Arnaud Quette <ArnaudQuette@eaton.com>


jimklimov:
Revised with Arno that this is not the place to apply this fix (the mods done for re-configuring into the nginx facade in one use-case are the place). See also comments in #716.

PR #716, opened 2020-05-19
Fix listening interface, set to localhost

aquette:
Signed-off-by: Arnaud Quette <ArnaudQuette@eaton.com>
(cherry picked from commit d8620c25a84adfcee2da380add20697d7e5062a3)

### DIFF
--- a/src/web/tntnet.xml
+++ b/src/web/tntnet.xml
@@ -614,11 +614,13 @@
   <!-- listen to a port (empty string means any local interface) -->
   <listeners>
     <listener>
+      <ip>127.0.0.1</ip>
       <port>8000</port>
     </listener>
   <!-- a certificate enables ssl -->
   <!-- ssl start
     <listener>
+      <ip>127.0.0.1</ip>
       <port>8443</port>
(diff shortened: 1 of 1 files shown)


src/web/tntnet.xml
```diff
@@ -614,11 +614,13 @@
   <!-- listen to a port (empty string means any local interface) -->
   <listeners>
     <listener>
+      <ip>127.0.0.1</ip>
```
EmilienKia, line 617: Will it listen on ::1 also ?

jimklimov:
Revised with Arno that this is not the place to apply this fix (the mods done for re-configuring into the nginx facade in one use-case are the place), although the question on `::1` from Emilien is a good one. I don't know. We didn't really support IPv6, might even actively avoid it in some places (maybe largely due to `localhost` often resolving to `either 127.0.0.1 or ::1` and one of those not being served by this or that daemon config). I suppose making sure everything works with IPv6 consistently would be a very separate endeavour, and here we need to ensure there would be no surprises for IPv4 at least ;)

PR #717, opened 2020-07-08
[PQSWPRG-7543] auth_verify.ecpp : pretend user.reauth(true) if session is OK…

jimklimov:
… and this lowest-priority auth implem ends up being THE handler

We have code elsewhere (secw, certgen) whose REST API wants sensitive operations re-authenticated every time. However the default auth implementation in fty-rest does not have such a concept and logic; third-party extended user management can provide that. For open-source builds to be usable, short-circuit by default the "user was (re)authenticated" flag during request processing, if the session is valid.

This is a no-op if a higher-priority auth implementation is used in a different bundling of components instead of this file.

### DIFF
--- a/src/web/src/auth_verify.ecpp
+++ b/src/web/src/auth_verify.ecpp
@@ -69,6 +69,13 @@ if(!checked_access_token.empty()) {
         user.gid (gid);
         user.login (user_name);
         delete[] user_name; user_name = NULL;
+
+        // Some of our code expects re-authentication for sensitive requests,
+        // if implemented by a third-party extended authentication module that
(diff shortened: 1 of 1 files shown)


jimklimov:
Seems the evaluation that this is a no-op if higher priority auth verifiers are present was wrong, and this servlet does get called after all. So this hardcoding of "reauth happened" should happen either conditionally, based on some system setting or detected presence of those other verifiers, or earlier in stack / per default setting of the flag, and flipped by others if deemed fit. Maybe a tri-state value, for definitive knowledge about flag state, is better than a bool here.

jimklimov:
Will revert in master for now.

PR #718, opened 2020-07-09
Fix of asset element ordering in the typology

perrettecl:


src/web/src/topology_location.ecpp
```diff
@@ -267,8 +343,20 @@ bool database_ready;
     }
     CLEANUP;
     #undef CLEANUP
+    
+    //order the data
+    cxxtools::SerializationInfo si;
+    
+    std::stringstream input;
+    input << json;
+    cxxtools::JsonDeserializer deserializer(input);
```
jeffdegott, line 352: shall be executed in a try catch block statement

PR #721, opened 2020-08-21
Modify static files regex pattern

boricj:
This is expected to be required fairly soon because of UI developments.

### DIFF
--- a/src/web/tntnet.xml
+++ b/src/web/tntnet.xml
@@ -40,7 +40,7 @@
     <mapping>
       <target>static@tntnet</target>
       <pathinfo>./$1</pathinfo>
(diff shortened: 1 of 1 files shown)


jimklimov:
Looks ok. On a more global note, is there a reason (e.g. auth/security) to serve UI assets via tntnet and not intercept these urls in nginx to serve there for IPM2 deployments? That RFE was posted a while ago ;)

jimklimov:
Also, the regex as is I think also matches /api calls leading to needless FS stats (can verofy by using DEBUG level logging in tntnet.xml on some deployment),  adding a negative lookahead to not match `^/api/.*` could be useful.

PR #727, opened 2020-12-02
[IPMVAL-2276] Split fty-rest tntnet.xml into better manageable pieces…

jimklimov:
… to replace asset mgmt implementation

NOTE: Wait for CI builds to complete, to check that the Makefile changes fit well.

Note also, that possibly in contrast to original request (the line numbering was along an undefined revision of tntnet.xml) at the time of this writing, the 51_asset_api-fty-rest.xml includes *all* matches of "/api/v1/asset*" URL handling (including "/api/v1/asset*s*"). If this is not the desired outcome, the list of files for sub-APIs can be split further.


src/Makemodule-local.am
```diff
@@ -319,30 +319,59 @@ install-exec-hook-cihelpers:
...
 	       -e 's|^\([[:blank:]]*\)ssl end -->|\1<!-- ssl end -->|' \
-	       -e 's|<!--\ <daemon>0</daemon>\ -->|<daemon>1</daemon>|' $< > $@ || \
-	rm -f $@
+	       -e 's|<!--\ <daemon>0</daemon>\ -->|<daemon>1</daemon>|' $(1) > $(2) \
+	|| rm -f $(2)
+endef
+
+### Most of the example files come unmodified
+%.xml.example: src/web/%.xml
+	cat < $< > $@
```
boricj, line 367: Why not `cp` here ?

boricj (approved):

src/Makemodule-local.am
jimklimov: No apparent reason, copy-pasta from other more complex XML concatenations... Does it impact much?

stark-dev (approved):

jimklimov:
For some reason CI complains about git_details generation which should not have been impacted by this branch. Checking what went fishy there now...

jimklimov:
@stark-dev @boricj : Seems the job generated from Jenkinsfile used by last master branch builds was still with old file, had build arguments and stuff. Generated jobs usually lag 1-2 builds after Jenkinsfile changes so that's maybe why.

The PR branch does not have build args at least, so possibly tries to run tests against a subdirectory if setup it never built.

jimklimov:
Phew, not my fault =D A rebuild of master branch has same deficiency as seen in the PR.

perrettecl (approved):

PR #728, opened 2020-12-03
Fix test build

stark-dev:
Signed-off-by: Mauro Guerrera <mauroguerrera@eaton.com>


etnKlendathu (approved):

jimklimov (approved):
I wonder if just hotfixing this is okay, or if the catch project should be updated. (And for good measure, a packaged one is better than a different copy in all sources).

But for a quick fix, LGTM :)

PR #736, opened 2021-01-13
auth_verify.ecpp improvement

jeffdegott:
Signed-off-by: F.-R. Degott <FrancoisRegisDegott@eaton.com>

### DIFF
--- a/src/web/src/auth_verify.ecpp
+++ b/src/web/src/auth_verify.ecpp
@@ -44,76 +44,76 @@ std::string gaccess_token;
 </%request>
 <%cpp>
 
-    // do nothing if user is yet authenticated
-    if (user.profile() != BiosProfile::Anonymous)
+    // handle not authenticated user only
+    if (user.profile() == BiosProfile::Anonymous)
     {
-        /* Go on to next module in tntnet.xml */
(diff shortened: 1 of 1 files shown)


jimklimov:
Hi, not sure about improvement - this seems like a rather suspicious change, although it is hard to evaluate the overall change vs. master ;)

At least one red flag is that with the new code structure, it dies for anonymous-profile access. The (original) verify servlet IIRC was to verify that the provided access token (if any) is valid and leads to a non-anonymous profile, but an anonymous profile without the token is IIRC also valid. Though looking at how this tests the `gaccess_token` against the regex-or-die (and did so in earlier branches), I'm not sure of that anymore.

Also this code structure skips a lot if the profile is not anonymous, so I think precludes re-authentication (e.g. needed for work with security wallet to reassure the server that you is still you and not somebody who took your logged-in browser for a ride).

While testing it in practice, please make sure that things intended to work without/before the login (e.g. the login page itself, minimal /admin/sysinfo report) and with re-auth still work.

And also note that while 42ity code ecosystem is used in the bigger product, there was an aim to have it usable on its own so contributors could build and use (dev-test) it standalone...

PR #750, opened 2021-03-09
src/Makemodule-local.am: use "ecppc -L" to avoid line macros that can…

jimklimov:
… cause ccache misses

It seems that these line macros embed a path relative to where the tool ran,
and if e.g. in distcheck we run in a .../_build/sub/ as current directory,
these paths starting with ./src/ do not lead anywhere useful (../src/ would).

Checking a theory from recent investigation in our CI farm. If this does end up speeding up the builds, this solution should be cloned to other *-rest components which use `ecppc`.

### DIFF
--- a/src/Makemodule-local.am
+++ b/src/Makemodule-local.am
@@ -141,7 +141,7 @@ package_CLIENTS_BIOS_CSV =    $(DESTDIR)$(bindir)/bios-csv
 ci_helperdir =         $(clientdir)
 
 ECPPC ?= ecppc
-ECPPFLAGS = --nolog
+ECPPFLAGS = --nolog -L
 ECPPFLAGS_CPP = -I$(top_builddir)/include -I$(top_srcdir)/include
 
 ECPPFILES= \


jimklimov:
Differences in generated `.cc` files are what the option help says -- absent `#line` macros (I wonder now what the downside could be? worse source-targeting in debugging messages emitted?):

````
$ diff -bu /tmp/hw_capability.cc.{with*,no*}
--- /tmp/hw_capability.cc.withLine      2021-03-09 12:38:55.652288565 +0000
+++ /tmp/hw_capability.cc.noLine        2021-03-09 12:39:41.731548195 +0000
@@ -14,7 +14,6 @@
 #include <stdexcept>

 // <%pre>
-#line 25 "./src/web/src/hw_capability.ecpp"

 #include <string>
 #include <fty_common_mlm_tntmlm.h>
@@ -157,11 +156,9 @@
  {
   tnt::DataChunks data(rawData);

-#line 117 "./src/web/src/hw_capability.ecpp"
   typedef UserInfo user_type;
   TNT_REQUEST_GLOBAL_VAR(user_type, user, "UserInfo user", ());   // <%request> UserInfo user
   // <%cpp>
-#line 119 "./src/web/src/hw_capability.ecpp"

     // permission check
     static const std::map <BiosProfile, std::string> PERMISSIONS = {
@@ -189,7 +186,6 @@
         http_die ("internal-error", errstr.c_str ());
     build_json (output, hw_type, gpi_count, gpo_count);

-#line 146 "./src/web/src/hw_capability.ecpp"
   reply.out() << ( output );
   reply.out() << data[0]; // \n
   // <%/cpp>
````

PR #751, opened 2021-03-09
src/Makemodule-local.am: use "ecppc -I" to avoid values in #line macros…

jimklimov:
…that can cause ccache misses

Exploring an alternative to #750, such which would not compromise fast builds vs debug-ability and troubleshooting of the result.

### DIFF
--- a/src/Makemodule-local.am
+++ b/src/Makemodule-local.am
@@ -143,13 +143,6 @@ ci_helperdir =         $(clientdir)
 ECPPC ?= ecppc
 ECPPFLAGS = --nolog
 ECPPFLAGS_CPP = -I$(top_builddir)/include -I$(top_srcdir)/include
-# By default eccpc embeds #line macros into generated .cc files to help the
-# debug messages pinpoint real culprits with __FILE__ and __LINE__ values.
-# These macros point to a relative path starting with "./src/" which is not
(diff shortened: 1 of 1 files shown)


jimklimov:
The `-I` with relative path values did not help, this more or less matches ccache docs. Retrying with absolutes.

jimklimov:
This approach does not seem to help: while `g++` launched by `make` has the absolute-path arguments (checked in /proc/NNN/cmdline as well as logged by ccache itself as "Command line:" just after "Config:" listing), the eventual `/usr/bin/g++` launched by `ccache` as the preprocessor and eventually as the "real compiler" has the arguments optimized and relativized. Not sure it would look at include dirs for `#line` macro dereferencing at all, but with resulting non-absolute dirs it certainly does not.

jimklimov:
Setting `hash_dir = false` also did not help, clean re-compilation of same stuff from another and then yet another symlink to workspace calls "real compiler" again.

jimklimov:
Botched the recent experiments, forgot to re-set CCACHE_BASEDIR (to the apparent workspace from which `make distcheck` is launched). With that fixed, seems ccache IS used with both enabled and disabled "hashdir" setting for the first run in a new apparent workspace.

PR #754, opened 2021-03-16
Use std::lock_guard

perrettecl:


jimklimov (approved):

src/shared/augtool.cc
```diff
@@ -100,12 +103,16 @@ augtool* augtool::get_instance() {
...
+    std::lock_guard<std::mutex> lock(in_mux);
+
     if(inst.prc == NULL) {
         MlmSubprocess::Argv exe = { "sudo", "augtool", "-S", "-I/usr/share/fty/lenses", "-e" };
         inst.prc = new MlmSubprocess::SubProcess(exe,
                                                  MlmSubprocess::SubProcess::STDOUT_PIPE |
                                                  MlmSubprocess::SubProcess::STDIN_PIPE);
+
+        //sleep to ensure augeas is launched
+        std::this_thread::sleep_for(std::chrono::seconds(2));
```
jimklimov, line 115: Just in case: is it a concern that to fork a process we always take a 2-second lag hit? Is there some better way to sync? (Maybe loop for up to 2 seconds in smaller sleeps, checking if it isRunning() to react to that as soon as it does run?)

And also what about grossly overloaded systems which at that moment exceed the 2 seconds (we did see "hrtimer" handling in excess of 1 second at least in the logs, for systems otherwise seeming to behave decently, so some thread blocking for too long especially on VM is a possibility).

jimklimov (approved):

src/shared/augtool.cc
aquette: as told to @perrettecl , the initial launch of augtool may lag a bit (due to the parsing of all lenses).
Hence, even 2sec may not be enough

jimklimov: ok, if that is the cause (I understood we do not even have the stdin/stdout pipe after too short a wait, or the process is not running).

Actually that long parsing is why there is a subprocess here on one hand, and we have a small selection in `/usr/share/fty/lenses` on another. With the complete standard collection of lenses it took IIRC 6-10 seconds to start on the IPCs.

Then the "what if" question about 2 sec not being enough (vs. dynamically deciding that we can proceed ASAP) may be not too rhetoric...

PR #759, opened 2021-03-30
PQSWMBT-4819 - getlog: add release-details.json in maintenance report

jeffdegott:
* getlog: add release-details.json in maintenance report
  Signed-off-by: Degott, Francois Regis <FrancoisRegisDegott@eaton.com>


aquette (approved):
Side, the manifest of the current version would be good to add too. We may need a symlink "current" to ease the listing though

PR #769, opened 2021-05-07
PQSWMBT-4875 - add firmwares info in maintenance report

jeffdegott:
firmwares.cfg (mass-management firmwares) included in maintenance report


aquette:
2 questions:
* is it really for release/IPM-2.3.0 or for 2.3.1? if it's the latter, please retarged
* what's the behavior if the file (/var/lib/fty/etn-mass-management/firmwares.cfg) doesn't exist?

PR #772, opened 2021-06-07
Add configured option and counter for asset and alert list

eric3873:


src/web/src/assets_in.ecpp
```diff
@@ -137,6 +138,8 @@ bool database_ready;
     std::string without = "";
     std::string status = "";
     std::string details = "";
+    std::string configured = "";
+    std::vector<std::string> checked_configured = { "all", "no", "yes" };
```
jeffdegott, line 142: empty value is not allowed?  or 'all' eq. empty?  (the common-db api don't handled 'all' value for the configured arg)

src/web/src/assets_in_counter.ecpp
```diff
@@ -0,0 +1,211 @@
...
+    catch (const std::exception& e) {
+        log_error ("Exception caught: '%s'.", e.what ());
+        std::string err = JSONIFY (e.what ());
+        http_die ("internal-error", err.c_str ());
+    }
+    // do the stuff
+    int counter = 0;
+    try {
+        counter = assets_in_count_container(connection, checked_id, checked_types, checked_subtypes, checked_capabilities, without, status, configured);
+    } catch (std::exception &e) {
```
jeffdegott, line 199: catch (const ...

jeffdegott:
Remember to update the "RFC11 Rest api" documentation on Confluence

src/web/src/assets_in.ecpp
eric3873: yes the "all" value (like empty) deactivate the option

src/web/src/assets_in_counter.ecpp
eric3873: Take into account

PR #779, opened 2021-06-30
Fix outlets list

eric3873:


src/db/inout/exportcsv.cc
```diff
@@ -612,8 +641,15 @@ void export_asset_json(std::ostream& out, std::set<std::string>* listElements)
 
                 cxxtools::SerializationInfo& si_outlet_label = si_outlet_x.addMember("");
                 si_outlet_label.addMember("name") <<= "label";
-                si_outlet_label.addMember("value") <<= oneOutlet.second.label;
-                si_outlet_label.addMember("read_only") <<= oneOutlet.second.label_r;
+                // if label is empty, get index value
```
jeffdegott, line 644: comment shall be updated

jeffdegott (approved):

src/db/inout/exportcsv.cc
aquette: comment still seem to be valid. Though I may stand corrected by missing something

jeffdegott: ok

PR #783, opened 2021-07-28
Fix issue with status.battery and status.battery.charger values

eric3873:


ecpp/current.ecpp
```diff
@@ -362,7 +372,20 @@ std::vector<uint32_t> asset_ids;
 
             if( one_measurement.first == "status.ups" )
             {
-              siJson.addMember(one_measurement.first) <<= shared::upsstatus_to_string(uint16_t(one_measurement.second));
+                siJson.addMember(one_measurement.first) <<= shared::upsstatus_to_string(uint16_t(one_measurement.second));
+            }
+            // special treatment for battery status and battery charger status and battery alarm code (string values)
+            else if( one_measurement.first == "status.battery" || one_measurement.first == "status.battery.charger" ||
```
jeffdegott, line 378: instead of (re)listing metric exceptions here, simply code at line 372 (before if(status.ups)):
search .first (metricName) in str_measurements map... if found, addMember in siJson and continue....

ecpp/current.ecpp
```diff
@@ -222,25 +222,35 @@ std::vector<uint32_t> asset_ids;
...
+            }
+            if(mresult.size() == 0) {
+                continue;
+            }
+            for(auto &data1 : mresult) {
+                double dvalue = 0.0;
+                std::string item_name = fty_proto_type (data1);
+                // TODO: non double values are not (yet) supported
+                // Special treatment for battery status and battery charger status and battery alarm code (string values)
+                if( item_name == "status.battery" || item_name == "status.battery.charger" || item_name == "battery.alarm.code") {
```
jeffdegott, line 240: "operatingStatus", "status.ups" exceptions can be safely handled here using str_measurements map...

eric3873: Not possible because values are numerics

ecpp/current.ecpp
eric3873: Take into account

PR #788, opened 2021-09-03
IPMPROG-325 - current.ecpp reviewed

jeffdegott:
* ecpp/current.ecpp review
Signed-off-by: Degott, Francois Regis <FrancoisRegisDegott@eaton.com>

### DIFF
--- a/ecpp/current.ecpp
+++ b/ecpp/current.ecpp
@@ -218,7 +218,8 @@ bool database_ready;
             const std::vector<std::string> strQuantities{
                 "status.battery",
                 "status.battery.charger",
-                "battery.alarm.code"
+                "battery.alarm.code",
+                "power.status"
             };
 
             for (auto &metric : metrics) {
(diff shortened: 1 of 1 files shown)


ecpp/current.ecpp
```diff
@@ -26,43 +26,40 @@
...
-            {7, "Dormant"},
-            {8, "Completed"},
-            {9, "Migrating"},
-            {10, "Immigrating"},
-            {11, "Emigrating"},
-            {12, "Snapshotting"},
-            {13, "In_Test"},
-            {14, "Transitioning"},
-            {15, "In_Service"}
+        {1, "None"},
```
aquette, line 48: this can probably be inherited from fty-service-status (https://github.com/42ity/fty-service-status/blob/master/include/fty_service_status.h#L37) though I don't quickly see status strings there

aquette (approved):

ecpp/current.ecpp
stark-dev: Indeed, those constants are defined here:
https://github.com/42ity/fty-service-status-systemd/blob/master/src/systemd.cpp#L29

Example of usage here https://github.com/42ity/fty-service-status-systemd/blob/master/example/src/example.cpp#L49

PR #792, opened 2021-09-16
Fix bad initialization of variable in timegm function and add mutex 

eric3873:
NOTE: TESTS DESACTIVATED FOR THIS PART. UNIT TESTS HAVE BEING DESACTIVATED DURING CMAKE MIGRATION ???

### DIFF
--- a/src/shared/utils.cc
+++ b/src/shared/utils.cc
@@ -23,6 +23,8 @@
 #include <assert.h>
 #include <fty_common.h>
 
+std::mutex timegm_mux;  // Mutex for my_timegm function which is not thread-safe
+
 bool is_average_step_supported(const char* step)
 {
     if (!step) {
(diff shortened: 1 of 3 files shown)


aquette (approved):

perrettecl (approved):

src/shared/utils.cc
```diff
@@ -22,6 +22,9 @@
 #include "cleanup.h"
 #include <assert.h>
 #include <fty_common.h>
+#include <mutex>
+
+std::mutex timegm_mux;  // Mutex for my_timegm function which is not thread-safe
```
jeffdegott, line 27: shall be declared static in my_timegm()

eric3873: Thanks you are right it could be, I will changed it if the PR is reopened

PR #814, opened 2022-01-23
[IPMPROG-2701] ecpp/license_POST.ecpp: let start-db-services run…

jimklimov:
…as long as it needs to ensure EULA wizard is handled well

Currently the script is aborted soon after start (due to 15 sec wait and then kill) while the db-init and other backend services take much longer to complete the start-up and to handle the rest of EULA wizard and other activities (approx 90 sec in local-VM testing; generally depends on hypervisor and its storage congestion vs. peak performance).

Notably, the early abort of the script by tntnet precludes starting the "ipm2-target-watchdog" after all those initializations, so the system does not recover if some units fail later (they stop and restart, but units who depended on them stop and have no cause to start... watchdog timers try to ensure they do so at least at some point; in newer systemd we could use Upholds=... for better effect).

### DIFF
--- a/ecpp/license_POST.ecpp
+++ b/ecpp/license_POST.ecpp
@@ -164,146 +164,158 @@ bool database_ready;
         // the wait() would end earlier if the child process finishes:
         int wait_cycle = 3 * 60000;
 
-        // configured in seconds
+        // configured in seconds, waiting in milliseconds:
         unsigned int mrt = tnt::TntConfig::it().maxRequestTime;
         if (mrt > 0 && mrt < (INT_MAX / 1000)) { // if enabled at all
             wait_cycle = static_cast<int>(mrt) * 666 ; // convert to msec and trim 2/3 for overheads
         }
         log_debug ("Starting db service will refresh tntnet deadman timer every %u msec", wait_cycle);
 
-repeat_start_db_services:
-        auto rv_svc = proc.wait(wait_cycle);
-        if (!rv_svc && "timeout" == rv_svc.error()) {
-            // Alas, "Expected" class deletes the assignment etc. operators,
-            // so we can not stash a value or change a variable, and have to
-            // loop old-school way with goto...
+        for (;;) { // repeat_start_db_services:
+            // May finish before "wait_cycle" elapses:
+            auto rv_svc = proc.wait(wait_cycle);
             uint64_t tme_finish = uint64_t(::time (NULL));
-            log_info ("Starting db service takes longer than expected on this system, waiting more, timestamp=%" PRIu64 " (duration=%" PRIu64 ")" ,
-                tme_finish, (tme_finish - tme_start));
-            request.touch();
-            goto repeat_start_db_services;
-        }
-        proc_out = proc.readAllStandardOutput();
-        proc_err = proc.readAllStandardError();
-        proc.kill();
 
-        uint64_t tme_finish = uint64_t(::time (NULL));
-        if (rv_svc) {
-            log_info ("Starting db service completed, timestamp=%" PRIu64 " (duration=%" PRIu64 "), exit-code=%i ..." ,
-                tme_finish, (tme_finish - tme_start), *rv_svc);
-        } else {
-            log_info ("Starting db service completed, timestamp=%" PRIu64 " (duration=%" PRIu64 "), exit-code=%s ..." ,
-                tme_finish, (tme_finish - tme_start), rv_svc.error().c_str());
-        }
-        if (!rv_svc || *rv_svc != 0) {
-            log_error ("Starting of start-db-services have failed. Consult system logs");
-            log_error ("%s failed with error %s.\n===== STDOUT: =====\n%s\n===== STDERR: =====\n%s\n=====",
-                "/usr/libexec/fty/start-db-services", rv_svc.error().c_str(), proc_out.c_str (), proc_err.c_str ());
-
-            if (rv_svc.error() != "timeout") {
-                std::string err =  TRANSLATE_ME ("Starting of start-db-services have failed. Consult system logs");
-                log_error_audit ("Request CREATE license FAILED");
-                http_die ("internal-error", err.c_str ());
-            }
-            // Negative return means killed by signal, typically by our
-            // subprocess timeout handling - fall through to try using DB
-
-            {
-                std::string proc_out1, proc_err1;
-                int lastResult = 0;
-                for (int i = 0; i < 10; ++i) {
-                    fty::Process::Arguments proc_cmd = {"systemctl", "list-ipm-units", "--active", "fty-db-engine"};
-                    // This should get our "systemctl" wrapper via PATH
-
-                    auto rv = fty::Process::run("sudo", proc_cmd, proc_out1, proc_err1);
-
-                    if (!rv) {
-                        log_error(rv.error().c_str());
-                        log_error_audit ("Request CREATE license FAILED");
-                        http_die ("internal-error", TRANSLATE_ME ("Database software service is not running").c_str());
-                    }
-
-                    lastResult = *rv;
-                    if (*rv == 0) {
-                        break;
-                    }
-                    sleep(5);
-                }
-
-                if (lastResult != 0) {
-                    std::string message;
-                    message = "`sudo systemctl list-ipm-units --active fty-db-engine`"
-                        " failed (service is not currently active). Return value = '"
-                        + std::to_string (lastResult) + "', stderr = '" + proc_err1 + "'." ;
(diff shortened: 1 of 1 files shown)


ecpp/license_POST.ecpp
```diff
@@ -142,7 +144,45 @@ bool database_ready;
...
+        // processing loop. It should suffice to catch `unexpected("timeout")`
+        // (child still running as of last check before time limit; no other
+        // errors) as per fty-utils::fty/process.h wording of Process::wait().
+        // the wait() would end earlier if the child process finishes:
+        int wait_cycle = 3 * 60000;
+
+        // configured in seconds
+        unsigned int mrt = tnt::TntConfig::it().maxRequestTime;
+        if (mrt > 0 && mrt < (INT_MAX / 1000)) { // if enabled at all
+            wait_cycle = static_cast<int>(mrt) * 666 ; // convert to msec and trim 2/3 for overheads
```
etnKlendathu, line 183: SAtan is watching you?

ecpp/license_POST.ecpp
```diff
@@ -142,7 +144,45 @@ bool database_ready;
...
+        auto rv_svc = proc.wait(wait_cycle);
+        if (!rv_svc && "timeout" == rv_svc.error()) {
+            // Alas, "Expected" class deletes the assignment etc. operators,
+            // so we can not stash a value or change a variable, and have to
+            // loop old-school way with goto...
+            uint64_t tme_finish = uint64_t(::time (NULL));
+            log_info ("Starting db service takes longer than expected on this system, waiting more, timestamp=%" PRIu64 " (duration=%" PRIu64 ")" ,
+                tme_finish, (tme_finish - tme_start));
+            request.touch();
+            goto repeat_start_db_services;
```
etnKlendathu: What? Goto? no!!!!! use while please

ecpp/license_POST.ecpp
```diff
@@ -142,7 +144,45 @@ bool database_ready;
...
+        log_debug ("Starting db service will refresh tntnet deadman timer every %u msec", wait_cycle);
+
+repeat_start_db_services:
+        auto rv_svc = proc.wait(wait_cycle);
+        if (!rv_svc && "timeout" == rv_svc.error()) {
+            // Alas, "Expected" class deletes the assignment etc. operators,
+            // so we can not stash a value or change a variable, and have to
+            // loop old-school way with goto...
+            uint64_t tme_finish = uint64_t(::time (NULL));
+            log_info ("Starting db service takes longer than expected on this system, waiting more, timestamp=%" PRIu64 " (duration=%" PRIu64 ")" ,
```
etnKlendathu: better to use new notation

ecpp/license_POST.ecpp
```diff
@@ -142,7 +144,45 @@ bool database_ready;
...
+            wait_cycle = static_cast<int>(mrt) * 666 ; // convert to msec and trim 2/3 for overheads
+        }
+        log_debug ("Starting db service will refresh tntnet deadman timer every %u msec", wait_cycle);
+
+repeat_start_db_services:
+        auto rv_svc = proc.wait(wait_cycle);
+        if (!rv_svc && "timeout" == rv_svc.error()) {
+            // Alas, "Expected" class deletes the assignment etc. operators,
+            // so we can not stash a value or change a variable, and have to
+            // loop old-school way with goto...
```
etnKlendathu: Very strange reason for use goto... really. Why do you need to stash variable?

ecpp/license_POST.ecpp
jimklimov: Sure, 1000 * 2/3 in integer maths, to allow decent overhead between timing this and context switches and other activities :)

ecpp/license_POST.ecpp
jimklimov: can't inherit `auto rv_svc` result otherwise, as discussed nearby (lower in comments, or above in code)

ecpp/license_POST.ecpp
jimklimov: I just copy-pasted existing precedent from nearby lines so did not mix styles within the file.

ecpp/license_POST.ecpp
jimklimov: `rv_svc` is used lower (in lines 190-198 of this codebase revision) in the context of being an `int-or-errorMessageCarrier` and apparently I can not just save a copy by `auto rv_svc = rv_svc_tmp` while in a loop using the tmp variable that would indeed be more sane - meaning I tried but could not pre-define an `rv_svc` before a sane while-loop, and assign a value during the cycles (whether directly from method, or using intermediate variables). So had to go crazy way which just works.

Code of Expected class is peppered with `deleted` methods to forbid copying, assignment, etc. so I guess that is intentional and I did not address that part.

jimklimov: Thinking of it, maybe we can pre-define an `int` and a... `string`?.. to store the parsing result from the rv_svc interpretation during the loop, and later consider null-ness of the string as the lack of exception so handling the int side as process success/failure, or else treat it as an exception from Proc class. That can get rid us of the goto, in exchange for some non-idiomatic craziness elsewhere :)

etnKlendathu: just example.
```
for(;;) {
   auto rc = wait();
   if (!rc && rc.error() == timeout) {
      ....
      continue;
   } else (!rc) {
      return error;
   } else {
      all good...;
      brak or return;
   }
}
```

perrettecl (changes requested):
Cannot be merge on 2.4.0

ecpp/license_POST.ecpp
jimklimov: That would impact existing code a lot more than I wanted, but can try, thanks :)

ecpp/license_POST.ecpp
```diff
@@ -142,128 +144,178 @@ bool database_ready;
...
+            if (!rv_svc || *rv_svc != 0) {
+                log_error ("Starting of start-db-services have failed. Consult system logs");
+                log_error ("%s failed with error %s.\n===== STDOUT: =====\n%s\n===== STDERR: =====\n%s\n=====",
+                    "/usr/libexec/fty/start-db-services", rv_svc.error().c_str(), proc_out.c_str (), proc_err.c_str ());
 
-                    if (!rv) {
-                        log_error(rv.error().c_str());
-                        log_error_audit ("Request CREATE license FAILED");
-                        http_die ("internal-error", TRANSLATE_ME ("Database software service is not running").c_str());
+                if (rv_svc.error() != "timeout") {
```
zjes: Could be problem.
you allow to go into this scope by *rv_svc != 0... means this variable has value, not error. But in this line you check error of possible value.

jimklimov: Sorry, got no opinion about that. Just indented original code as it was. Just looking around, apparently the `Expected<int>` which that `auto rv_svc` is, works that way. Similarly, no idea why `auto` instead of specific types with expected behaviors.

jimklimov: Thanks for the catch, confirmed as problematic by killing the script (via terminal) while it was being processed. Fix pending, but workstation got no internet to post it at the moment.

ecpp/license_POST.ecpp
jimklimov: In later revisions, made this behavior optional. It turned out to be better to use shorter sleep cycles (to not allow blocking for long on child process `write()`s, and drain their buffers quickly and regularly).

ecpp/license_POST.ecpp
jimklimov: Not exactly like above, but the suggestion was fruitful, thanks.

jimklimov:
In later revisions, it turned out to be better to drain child process stdout/stderr buffers quickly and regularly and use shorter sleep cycles (to not allow blocking for long on child process `write()`s), so the logic for longer wait-before-kill with timeouts based on deadman config (e.g. a large portion of 10 minutes by default) seems currently useless, we should go for a few seconds between status checks (and buffer drains) at most.

Bigger problem was that the child process never ended after completing its logic and reporting the collected log - and here it turned out that the fty-utils process.h class lies in the names about `readAllStandardError()` and `readAllStandardOutput()` - in fact they only try two select's (by default arg of `readFromFd()`), so any bulkier reads remain queued, and the `write()` in scripted child processes remain blocked on full buffer. A series of reads until empty has solved this, as tested in practice: finally, the EULA wizard proceeds as soon as (and no sooner than) the `start-db-services` script logs its exit message.

jimklimov:
> Cannot be merge on 2.4.0

Retargeted to 2.5.0

perrettecl (approved):

PR #846, opened 2022-11-05
IPMPROG-4911 - fix smtp config double quote

FrancoisRegisDegott-eaton:
config.ecpp
The GET request won't work for smtp user/password config (the decoder is not applied)
1/ fix: set the correct property names to apply decode()
2/ added a 'smtp' parent property filter to exactly balance smtp user/passwd encode/decode process

### DIFF
--- a/ecpp/config.ecpp
+++ b/ecpp/config.ecpp
@@ -270,6 +270,12 @@ zconfig_remove (zconfig_t **self_p)
     *self_p = NULL;
 }
 
+zconfig_t* zconfig_parent (zconfig_t* self)
+{
+    return self ? self->parent : NULL;
+}
+
+// set zconfig_t* from cxxtools::SerializationInfo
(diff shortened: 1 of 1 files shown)


ecpp/config.ecpp
```diff
@@ -315,27 +323,34 @@ void operator<<=(zconfig_t* root, const cxxtools::SerializationInfo& si) {
...
             si.setNull();
         }
         else {
-            std::string decodedValue(value);
-            std::string configName(zconfig_name(root));
-            if( (configName == "BIOS_SMTP_PASSWD") || (configName == "BIOS_SMTP_USER") ) {
-                decodedValue = quotecodec::quoteDecode(value);
+            std::string value(zc_value);
+
+            // IPMPROG-4911: handle '"' in SMTP PASSWD/USER
```
ericclappier-eaton, line 336: Symmetric between serialization/deserialization?

sinanyvs (approved):

PR #859, opened 2023-11-21
Update control (add libcurl-dev dependency

FrancoisRegisDegott-eaton:
ecpp/getlog_GET.ecpp needs libcurl


FrancoisRegisDegott-eaton:
needs curl, not libcurl   (sorry)

